PR #21, opened 2020-01-09
Split inplace/not inplace transform

matozoid:
In my opinion you always know if you want to transform in place or not when writing your code. So when I remove the boolean and make two separate methods for transformChildren, it becomes more readable and usable in my opinion. One getting/setting on the existing nodes, the other is creating new objects.


ftomassetti:
From the point of the user it makes sense, however in this way we duplicate code. What if we have an implementation function (private) taking the flag and two public functions calling the implementation function with the flag set to true and false?

matozoid:
In my opinion the code is improved too, do take a look at the result, the flows are mostly different. The code that is duplicated is the walk-the-tree-is-it-a-collection-or-not code. I see that pattern a lot and it might be generalized? However, I didn't find a good way to do that right away. Maybe this is a problem for later.

ftomassetti:
+1 for me but there are conflicts with other changes, I would merge once solved

Issue #18, opened 2020-01-09
Reimplement Node.transform(operation: (Node) -> Node, inPlace: Boolean): Node


matozoid:
There's a TODO there.

matozoid:
Can't we just throw "transform" away?

matozoid:
Here's what I think right now:
mapTree -> should be named mapChildren
transformTree -> should be named transformChildren
transform -> transforms the whole tree. Can become the new transformTree and mapTree by calling transformChildren and mapChildren recursively.

That would also solve #32

ftomassetti:
what do we mean by children and what by tree? Does "children" refers only to the direct children, while "tree" refers to the whole sub tree?
So for example:
```
A -> B -> D
        C  -> E
             -> F
```

Where A contains B and C, D contains B, and C contains E and F.

In this scenario mapChildren/transformChildren, when called on A will touch only B and C, but not D, E, F. Transform instead will operate on a tree and therefore when called on "A" will touch A, B,C,D,E, and F. When called on C, it will touch C, E, and F.

Is that correct?
If this is the case I would call transform "transformTree", to avoid ambiguity

matozoid:
So we can have:
mapChildren and transformChildren.
mapTree and transformTree.

The bottom two will use the top two (if I can get that to work)

ftomassetti:
It sounds great to me. Thank you for working with me on this one!

PR #22, opened 2020-01-09
Reimplement transform

matozoid:
#18 - if I understand correctly, the normal `transform` is like `transformChildren` plus `transform(this)`.

If #21 is accepted, this will have to be updated. There will be `transform` and `transformInPlace`

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -13,10 +13,10 @@ import kotlin.reflect.full.primaryConstructor
  */
 private val <T : Node> T.containmentProperties: Collection<KProperty1<T, *>>
     get() = this.javaClass.kotlin.memberProperties
-            .filter { it.visibility == KVisibility.PUBLIC }
-            .filter { it.findAnnotation<Derived>() == null }
-            .filter { it.findAnnotation<Link>() == null }
-            .filter { it.name != "parent" }
+        .filter { it.visibility == KVisibility.PUBLIC }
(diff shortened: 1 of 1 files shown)


matozoid:
I seem to misunderstand transformChildren: does it only affect the direct children of the node? If so, where is the function that affects all descendants?

ftomassetti:
Yes, it seems to affect only the direct children.
I think there is note an operation to transform all descendants. I think one could build it using `transformChildren` in the operation, making it recursive

ftomassetti:
I think the initial use case was replacing a child in an AST which may be immutable

matozoid:
Well, I made #23.

So, what exactly should `transform` do?

ftomassetti:
Perhaps we should wait and merge the others PRs, as they would affect this one, correct?

matozoid:
Okay, only this one left. If you had to Javadoc this function, what would it say?

ftomassetti:
My Javadoc would say this:
```
Transform can be used to operate on a whole AST, transforming single nodes as desired.
Depending on the AST being modifiable or not the transformation can happen in place or not. In the former case the existing AST is modified directly, otherwise a new AST is generated to replace the previous one.
``

ftomassetti:
Is this one blocked by something I should do @matozoid ?

matozoid:
It's waiting for some kind of agreement or disagreement here: https://github.com/Strumenta/kolasu/issues/18 :-)

ftomassetti:
ok, then I will continue discussion in #18 (done :D)

matozoid:
@matozoid I need to finish this :-)

matozoid:
Okay, naming is now in order :-)

PR #24, opened 2020-01-10
Create a depth-first AST traversing sequence

matozoid:
This is actually really, really powerful. I can do an in-place full tree transform like this:

```kotlin
    project.descendants().forEach { node ->
        when (node) {
            is HexadecimalLiteralExpr -> {
                val value = parseInt(node.rawValue.substring(1), 16)
                node.replace(ConstantValueExpr(value, guessLiteralNumberType(value)))
            }
        }
    }
```

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Navigation.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Navigation.kt
@@ -4,11 +4,5 @@ package com.strumenta.kolasu.model
  * @return the nearest ancestor of this node that is an instance of klass.
  */
 fun <T : Node> Node.ancestor(klass: Class<T>): T? {
-    if (this.parent != null) {
-        if (klass.isInstance(this.parent)) {
-            return this.parent as T
-        }
-        return this.parent!!.ancestor(klass)
-    }
-    return null
+    return walkAncestors().filterIsInstance(klass).firstOrNull()
 }
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -1,6 +1,6 @@
 package com.strumenta.kolasu.model
 
-import java.util.LinkedList
+import java.util.*
 import kotlin.collections.HashMap
 import kotlin.reflect.*
 import kotlin.reflect.full.findAnnotation
@@ -11,12 +11,12 @@ import kotlin.reflect.full.primaryConstructor
 /**
  * @return all properties of this node that are considered AST properties.
  */
-private val <T : Node> T.containmentProperties: Collection<KProperty1<T, *>>
+internal val <T : Node> T.containmentProperties: Collection<KProperty1<T, *>>
     get() = this.javaClass.kotlin.memberProperties
-        .filter { it.visibility == KVisibility.PUBLIC }
-        .filter { it.findAnnotation<Derived>() == null }
-        .filter { it.findAnnotation<Link>() == null }
-        .filter { it.name != "parent" }
+            .filter { it.visibility == KVisibility.PUBLIC }
+            .filter { it.findAnnotation<Derived>() == null }
+            .filter { it.findAnnotation<Link>() == null }
+            .filter { it.name != "parent" }
 
 /**
  * Sets or corrects the parent of all AST nodes.
@@ -24,7 +24,7 @@ private val <T : Node> T.containmentProperties: Collection<KProperty1<T, *>>
  * so this function should be called manually after modifying the AST.
  */
 fun Node.assignParents() {
-    this.children.forEach {
+    this.walkChildren().forEach {
         it.parent = this
         it.assignParents()
     }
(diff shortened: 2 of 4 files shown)


matozoid:
I've made this a "val" instead of a "fun" because all my Java years have taught me that getters should be simple retrievals of a value, not complex functions.

matozoid:
I've made it a fun instead of a val because my years with Java taught me that getters should just return a value. That's not in the same style as `containmentProperties` and `children` though.

matozoid:
I'm working on replacing existing code to use this, and am realizing that...
- the first method to build should be `walk(algorithm)`
- `walk` should output the start node too. We can always filter it away, but we cannot re-add it in the right position because the algorithm will decide where it goes.

matozoid:
Implemented that too. Note how some of the existing utility methods have turned into simple one-liners this way. Now you can map, filter, and foreach any order of nodes to whatever you want.

matozoid:
Okay, and here is the post-order leaves-first algorithm. Pretty huge and ugly but can't get it smaller right now.

src/main/kotlin/com/strumenta/kolasu/model/Navigation.kt
```diff
@@ -4,11 +4,5 @@ package com.strumenta.kolasu.model
...
  */
 fun <T : Node> Node.ancestor(klass: Class<T>): T? {
-    if (this.parent != null) {
-        if (klass.isInstance(this.parent)) {
-            return this.parent as T
-        }
-        return this.parent!!.ancestor(klass)
-    }
-    return null
+    return walkAncestors().filterIsInstance(klass).firstOrNull()
```
ftomassetti: This looks nice!

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -11,20 +11,20 @@ import kotlin.reflect.full.primaryConstructor
...
+            .filter { it.name != "parent" }
 
 /**
  * Sets or corrects the parent of all AST nodes.
  * Kolasu does not see set/add/delete operations on the AST nodes,
  * so this function should be called manually after modifying the AST.
  */
 fun Node.assignParents() {
-    this.children.forEach {
+    this.walkChildren().forEach {
```
ftomassetti: nice too

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -33,15 +33,8 @@ fun Node.assignParents() {
...
-fun Node.processNodes(operation: (Node) -> Unit) {
-    operation(this)
-    containmentProperties.forEach { p ->
-        val v = p.get(this)
-        when (v) {
-            is Node -> v.processNodes(operation)
-            is Collection<*> -> v.forEach { (it as? Node)?.processNodes(operation) }
-        }
-    }
+fun Node.processNodes(operation: (Node) -> Unit, walker: KFunction1<Node, Sequence<Node>> = Node::walk) {
```
ftomassetti: Should we add a comment to explain what walker does? I assume it cotnrols the order in which we traverse nodes

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -117,49 +110,22 @@ fun Node.processProperties(
...
-                    val res = it.find(predicate)
-                    if (res != null) {
-                        return res
-                    }
-                }
-            }
-        }
-    }
-    return null
+fun Node.find(predicate: (Node) -> Boolean, walker: KFunction1<Node, Sequence<Node>> = Node::walk): Node? {
```
ftomassetti: Another one made nicer and more readable!
I would comment walker also here, as it is not super intuitive

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -178,7 +144,7 @@ fun Node.processConsideringParent(operation: (Node, Node?) -> Unit, parent: Node
 }
 
 /**
- * @return all descendants of this node, meaning all the children, the children of those children, etc.
+ * @return all direct children of this node.
```
ftomassetti: So here the doc was stating a lie!

src/main/kotlin/com/strumenta/kolasu/model/Traversing.kt
```diff
@@ -0,0 +1,124 @@
+package com.strumenta.kolasu.model
+
+import java.util.*
+
+/**
+ * Some Kotlinization of Deques used as a stack.
+ */
+typealias Stack<T> = Deque<T>
```
ftomassetti: Nice touch

src/main/kotlin/com/strumenta/kolasu/model/Traversing.kt
```diff
@@ -0,0 +1,124 @@
...
+
+fun <T> mutableStackOf(): Stack<T> = ArrayDeque()
+fun <T> Stack<T>.pushAll(elements: Collection<T>) {
```
ftomassetti: nitpick: I would put a blank link between mutableStackOf and pushAll

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Too bad it was not caught in the review ;-)

src/main/kotlin/com/strumenta/kolasu/model/Traversing.kt
matozoid: Not sure how nice - it has the same name as the Java Stack class, and it exposes a bunch of non-stack methods. It shows intention better though. We can always throw it out again.

src/main/kotlin/com/strumenta/kolasu/model/Navigation.kt
matozoid: Use Kotlin's library to the max :-)

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Done

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Also done.

src/main/kotlin/com/strumenta/kolasu/model/Traversing.kt
matozoid: Ok

matozoid:
Woohoo!

PR #25, opened 2020-01-10
Use more Kotlin-like collections

matozoid:
If I understand the manual correctly, this is the way to go for collections in Kotlin.


ftomassetti:
+1

PR #28, opened 2020-01-12
Support only list

### LINKED ISSUE #19: Implement other collections in Node.transformChildren or throw exception

matozoid:
Fixes #19. Maybe not a great solution, but better than a TODO :-)

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -228,7 +228,7 @@ fun Node.transformChildrenInPlace(operation: (Node) -> Node) {
                         }
                     }
                 } else {
-                    TODO()
+                    throw UnsupportedOperationException("Only modifications in a List and MutableList are supported, not " + value::class.toString())
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -229,7 +229,7 @@ fun Node.transformChildrenInPlace(operation: (Node) -> Node) {
                         }
                     }
                 } else {
-                    TODO()
+                    throw UnsupportedOperationException("Only modifications in a List and MutableList are supported, not " + value::class.toString())
```
ftomassetti:
```suggestion
                    throw UnsupportedOperationException("Only modifications in a List and MutableList are supported, not " + value::class.toString())
```

```suggestion
                    throw UnsupportedOperationException("Only modifications in a List and MutableList are supported, not ${value::class.toString()}")
```

Kotlin has string interpolation :)

matozoid: Stupid Intellij didn't suggest it, so I was lazy :-)

matozoid: Aaaand I forgot to do this :-D

ftomassetti: I changed it and merged it

PR #30, opened 2020-01-14
Some naming suggestions

matozoid:
These are suggestions to make function names more expressive. I assume @ftomassetti is not going to agree with all of them, so let me know :-)


ftomassetti:
The only one I am not convinced about is the debugPrint. But multilineString is not great either, so...

matozoid:
That's exactly the one I wasn't sure about either :-) I was thinking about `dumpAst` in good old JP 2.0 tradition ;-)

PR #33, opened 2020-01-15
add replaceWithSeveral

matozoid:
I'm not sure the argument should be a list. Maybe a vararg. Not sure how this will be used in the future.

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -303,6 +303,7 @@ fun Node.replaceWithSeveral(oldNode: Node, newNodes: List<Node>) {
             }
         }
     }
+    throw IllegalStateException("Did not find $oldNode in any MutableList in $this. Replace failed.")
 }
 
 /**
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -29,7 +29,7 @@ fun Node.assignParents() {
 }
 
 /**
- * Recursively execute "operation" on this node, and all nodes below this node.
+ * Recursively execute [operation] on [this] node, and all nodes below this node.
  * @param walker the function that generates the nodes to operate on in the desired sequence.
  */
 fun Node.processNodes(operation: (Node) -> Unit, walker: KFunction1<Node, Sequence<Node>> = Node::walk) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -61,14 +61,14 @@ private fun providesNodes(kclass: KClass<*>?): Boolean {
 }
 
 /**
- * @return can this class be considered an AST node?
+ * @return can [this] class be considered an AST node?
```
ftomassetti: why the square brackets?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -29,7 +29,7 @@ fun Node.assignParents() {
 }
 
 /**
- * Recursively execute "operation" on this node, and all nodes below this node.
+ * Recursively execute [operation] on [this] node, and all nodes below this node.
```
ftomassetti: why the square brackets?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -272,12 +272,45 @@ fun Node.mapTree(operation: (Node) -> Node): Node {
...
         throw IllegalStateException("Parent not set")
     }
     this.parent!!.transformTree { if (it == this) other else it }
 }
+
+/**
+ * Looks for [oldNode] in the lists of nodes in this node.
+ * When found, it is removed, and in its place the [newNodes] are inserted.
+ */
+@Suppress("UNCHECKED_CAST") // assumption: a MutableList with a Node in it is a MutableList<Node>
```
ftomassetti: What happens if the property allows for one element and not a list?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -272,12 +272,45 @@ fun Node.mapTree(operation: (Node) -> Node): Node {
...
+fun Node.replaceWithSeveral(oldNode: Node, newNodes: List<Node>) {
+    relevantMemberProperties().forEach { property ->
+        when (val value = property.get(this)) {
+            is MutableList<*> -> {
```
ftomassetti: what if the value is the single node we were looking for? Should we throw an exception?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: It's a link, check out https://kotlinlang.org/docs/reference/kotlin-doc.html

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Nothing. Is that a valid use case or shall we make it an error?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Oh, we're asking each other the same thing :-)

matozoid: For my use case it's an error, so I'll make it an error.

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
ftomassetti: I would expect it to throw an error, it would be more clear to me

matozoid:
Is this okay, @ftomassetti ?

ftomassetti:
Looks great, thanks!

PR #35, opened 2020-01-19
Insert before after remove

matozoid:
This builds on #33, adding a few functions that make manipulation of lists of nodes complete as far as I can see.

There are sets of...
- functions that say "remove/addbefore/etc. THIS node" which use:
- functions that say "remove/addbefore/etc. the parameter node, which is somewhere in a list in THIS"
- functions that say "remove/addbefore/etc. the parameter item from THIS MutableList"

Possibly the second set of functions is redundant or confusing.

The function that looks for the node in all of the mutable lists in a node has been generalized.

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -303,6 +303,7 @@ fun Node.replaceWithSeveral(oldNode: Node, newNodes: List<Node>) {
             }
         }
     }
+    throw IllegalStateException("Did not find $oldNode in any MutableList in $this. Replace failed.")
 }
 
 /**
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -29,7 +29,7 @@ fun Node.assignParents() {
 }
 
 /**
- * Recursively execute "operation" on this node, and all nodes below this node.
+ * Recursively execute [operation] on [this] node, and all nodes below this node.
  * @param walker the function that generates the nodes to operate on in the desired sequence.
  */
 fun Node.processNodes(operation: (Node) -> Unit, walker: KFunction1<Node, Sequence<Node>> = Node::walk) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -272,12 +272,133 @@ fun Node.mapTree(operation: (Node) -> Node): Node {
...
         throw IllegalStateException("Parent not set")
     }
     this.parent!!.transformTree { if (it == this) other else it }
 }
+
+/**
+ * Looks for [oldNode] in the lists of nodes in this node.
+ * When found, it is removed, and in its place the [newNodes] are inserted.
+ */
+fun Node.replaceWithSeveral(oldNode: Node, newNodes: List<Node>) {
```
ftomassetti: Should we specify that in case a node is not contained in a list nothing will happen?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -272,12 +272,133 @@ fun Node.mapTree(operation: (Node) -> Node): Node {
...
+        oldNode.parent = null
+        newNodes.forEach { node -> node.parent = this }
+    }
+}
+
+/**
+ * Looks for [targetNode] in the lists of nodes in this node.
+ * When found, it is removed.
+ */
+fun Node.remove(targetNode: Node) {
```
ftomassetti: But this does not remove the node if it is not in a list, right? Should we rename the method?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: If I'm not mistaken it will throw an exception. Shall I document that?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Same as the other - or do you think the method name is too general?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
ftomassetti: Yes, please

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
ftomassetti: Yes, in this case the name is a bit general as it will not remove the node if it is not contained in a list, right?

matozoid: Done, is the name better?

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: Done

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
ftomassetti: It is perfect, thank you!

ftomassetti:
Thank you for your work on this one! It is going to be quite useful to have

matozoid:
Only the transform methods left to do!

PR #39, opened 2020-07-23
Suggested namechanges

matozoid:
I've kind of forgotten what it was I was still going to do, but I bet it all comes back if I just start with a simple thing like this suggested name change.

I'm kind of worried that no unit test was touched for this.

### DIFF
--- a/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -162,7 +162,7 @@ val Node.children: List<Node>
     }
 
 // TODO reimplement using transformChildren
-fun Node.transform(operation: (Node) -> Node, inPlace: Boolean = false): Node {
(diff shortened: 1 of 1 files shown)


matozoid:
image

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -162,19 +162,19 @@ val Node.children: List<Node>
     }
 
 // TODO reimplement using transformChildren
```
ftomassetti, line 164: We should also update this comment now, I think

ftomassetti:
Thank you!

src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
matozoid: I'll do that in the next PR.

PR #42, opened 2020-11-19
Kolasu seriazable

lanarimarco:


build.gradle
```diff
@@ -34,7 +36,7 @@ repositories {
 
 def kolasuVersion = "0.2.3-SNAPSHOT"
 
-project.group   = "me.tomassetti"
+project.group   = "me.lanarimarco"
```
ftomassetti, line 39: This should be removed

build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.kotlin_version = '1.3.41'
+    ext.kotlin_version = '1.4.0'
```
ftomassetti, line 2: We could migrate to 1.4.10, as it is more recent

ftomassetti (commented):
Nodes built with Kolasu should be already serializable to JSON and XML. 
You can look in classes `JsonGenerator` and `XmlGenerator`

PR #49, opened 2021-08-04
CLI support. Track parsing times. Some refactoring and linting.

alessiostalla:


core/build.gradle
```diff
@@ -8,8 +8,13 @@ plugins {
     id "org.jetbrains.dokka"
 }
 
-java.sourceCompatibility = "$jvm_version"
-java.targetCompatibility = "$jvm_version"
+java {
+    sourceCompatibility = "$jvm_version"
+    targetCompatibility = "$jvm_version"
+    registerFeature('cli') {
```
ftomassetti, line 14: I have never seen this. What is this used for?

core/src/main/kotlin/com/strumenta/kolasu/parsing/ParseTree.kt
```diff
@@ -70,6 +70,7 @@ fun ParserRuleContext.processDescendantsAndErrors(
     }
 }
 
+@Deprecated("Use KolasuParser")
```
ftomassetti, line 73: Good point

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -167,7 +173,10 @@ abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext> {
 
     protected abstract fun createANTLRLexer(inputStream: InputStream): Lexer
     protected abstract fun createANTLRParser(tokenStream: TokenStream): P
-    protected abstract fun invokeRootRule(parser: P): C?
+    protected open fun invokeRootRule(parser: P): C? {
+        val entryPoint = parser::class.memberFunctions.find { it.name == parser.ruleNames[0] }
```
ftomassetti, line 177: nice idea

ftomassetti: We could add a comment on this method, to explain users when they should override this method

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -176,26 +185,28 @@ abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext> {
...
-        lexer.injectErrorCollectorInLexer(issues)
         val tokens = LinkedList<Token>()
-        do {
-            val t = lexer.nextToken()
-            if (t == null) {
-                break
-            } else {
-                if (!onlyFromDefaultChannel || t.channel == Token.DEFAULT_CHANNEL) {
-                    tokens.add(t)
+        val time = measureTimeMillis {
```
ftomassetti, line 189: Does measuring slow us down potentially?

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -253,26 +266,37 @@ abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext> {
...
     }
 
     fun parse(file: File, considerPosition: Boolean = true): ParsingResult<R> = parse(
         FileInputStream(file),
         considerPosition
     )
+
+    // For convenient use from Java
+    fun walk(node: Node) = node.walk()
+    @JvmOverloads
```
ftomassetti, line 294: I would add a blank line before the annotation

core/src/main/kotlin/com/strumenta/kolasu/parsing/coverage/CoverageListener.kt
```diff
@@ -107,7 +107,7 @@ open class CoverageListener(var parser: Parser? = null, val expandUncoveredPaths
         val path = pathStack.peek().followWith(element)
         return if (!paths.containsKey(path)) {
             paths[path] = false
-            if(expandUncoveredPaths) {
+            if (expandUncoveredPaths) {
```
ftomassetti, line 110: +1

ftomassetti (approved):
It looks good, but I confess I have no idea how the cli feature works. I need to study it

core/build.gradle
alessiostalla: To have the clikt library as an optional dependency in the generated POM. So that consumers of Kolasu who don't use it don't get the dependency.

core/src/main/kotlin/com/strumenta/kolasu/parsing/ParseTree.kt
alessiostalla: I wasn't sure to delete it but the IDE flags it as duplicate code

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: OK, makes sense

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: I highly doubt it. My only concern is that it's not available for all platforms.
They have experimental cross-platform time APIs but... they're experimental and subject to change.

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: OK

core/src/main/kotlin/com/strumenta/kolasu/parsing/coverage/CoverageListener.kt
alessiostalla: The linter was making the build fail :)

alessiostalla:
> It looks good, but I confess I have no idea how the cli feature works. I need to study it

I use it like this:

```
fun main(args: Array<String>) {
    ParserCLI(SASLanguage()).subcommands(ParserCLIEMF<SourceFile>()).main(args)
}
```

Then I can invoke the resulting jar file like this:
```
java -jar target/sas-parser-1.0.0-jar-with-dependencies.jar examples/covid-19-sas/ccf/COVID_19.sas emf covid-19.xmi 
```
Result:
```
Saving metamodel to /Users/alessio/projects/sas-parser/metamodel.xmi... Done (1171ms).
Parsing examples/covid-19-sas/ccf/COVID_19.sas... OK.
Done (135161ms of which 131724ms to build the AST).
The file has 2826 lines and 138441 characters. The parse tree has 33548 nodes (138441 tokens). The AST has 15134 nodes.
Saving AST to /Users/alessio/projects/sas-parser/covid-19.xmi... Done (855ms).
```
(Yes this particular file takes a lot. I have yet to profile it.)

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: To elaborate, measureTimeMillis only does a system.currentTimeMillis() at the start and one at the end of the block and returns the difference. It's not instrumenting the code or other invasive stuff.

core/build.gradle
ftomassetti: How does one specify that he wants it when specifying Kolasu as a dependency?

ftomassetti:
@alessiostalla to avoid confusion: the PR is approved, so feel free to merge it when you want

core/build.gradle
alessiostalla: I gather that to use it from Gradle you'd have to write something like
```
implementation("com.strumenta.kolasu:kolasu-core:1.x.y")
implementation("com.strumenta.kolasu:kolasu-core:1.x.y") {
        capabilities {
           requireCapability("com.strumenta.kolasu:kolasu-core-cli")
        }
    }
```
This explains it https://blog.gradle.org/optional-dependencies
If it's too complicated we could either make the dependency mandatory (consumers who don't want it can still exclude it) or make it the equivalent of Maven's "provided" which I don't know in Gradle (i.e. consumers who want it must know about it & add the specific dependencies to their project autonomously).

PR #54, opened 2021-09-14
Feature: issue severity

alessiostalla:


emf/multipkg.json
```diff
@@ -10,6 +10,15 @@
     "eSuperTypes" : [ {
       "eClass" : "http://www.eclipse.org/emf/2002/Ecore#//EClass",
       "$ref" : "https://strumenta.com/kolasu/v1#//ASTNode"
+    } ],
+    "eStructuralFeatures" : [ {
```
ftomassetti, line 14: This I think would not be necessary because this information is already represented in EMF as the EClass of the node

emf/multipkg.json
```diff
@@ -18,6 +27,15 @@
...
```
ftomassetti, line 31: Same here

emf/multipkg.json
```diff
@@ -92,6 +110,14 @@
         "$ref" : "/0/Statement"
       },
       "containment" : true
+    }, {
```
ftomassetti, line 113: And here

ftomassetti (approved):
I think the issue severity is a good addition, my only comments are about the role of nodeType, which we discussed outside this conversation

emf/multipkg.json
alessiostalla: Actually that file should have been ignored, my mistake

PR #67, opened 2022-03-23
Strumenta#66 Printing AST skips private list fields

loradd:


core/src/main/kotlin/com/strumenta/kolasu/model/Printing.kt
```diff
@@ -50,34 +50,36 @@ fun Node.debugPrint(indent: String = "", configuration: DebugPrintConfiguration
             } else {
                 val mt = property.returnType.javaType
                 if (mt is ParameterizedType && mt.rawType == List::class.java) {
-                    if (property.get(this) == null) {
-                        sb.append("$indent$indentBlock${property.name} = null")
-                    } else {
-                        val value = property.get(this) as List<*>
-                        if (value.isEmpty() && configuration.skipEmptyCollections) {
-                            // nothing to do
+                    if (property.visibility == PUBLIC) {
```
ftomassetti, line 53: Would it make sense to consider package-protected properties? If not, should we add a warning or write this somewhere?

ftomassetti (approved):
+1

core/src/main/kotlin/com/strumenta/kolasu/model/Printing.kt
loradd: What about handling both private, protected and internal properties according to some dedicated configuration parameter in `DebugPrintConfiguration`, e.g. `val skipProtectedProperties: Boolean = false`?

loradd: > Would it make sense to consider package-protected properties? If not, should we add a warning or write this somewhere?

What about handling both private, protected and internal properties according to some dedicated configuration parameter in `DebugPrintConfiguration`, e.g. `val skipProtectedProperties: Boolean = false`?

ftomassetti: I think that may be useful, yes

ftomassetti:
Thank you Lorenzo. Can you solve the conflicts and merge the PR? I think you should have the rights to do so

PR #69, opened 2022-04-11
Proposal for the decoupling of Node from ANTLR.

alessiostalla:
This is a breaking change. It could be part of Kolasu 2.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -17,6 +17,12 @@ open class Node(): WithPosition {
         }
     }
 
+    constructor(origin: WithPosition?) : this() {
+        if (origin != null) {
+            this.origin = origin
+        }
+    }
+
     open val nodeType: String
         get() = this.javaClass.canonicalName
 
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -4,12 +4,12 @@ interface WithPosition {
     val position: Position?
 }
 
-class JustPosition(override val position: Position): WithPosition
+class JustPosition(override val position: Position) : WithPosition
 
 /**
  * The Abstract Syntax Tree will be constituted by instances of Node.
  */
-open class Node(): WithPosition {
+open class Node() : WithPosition {
 
     constructor(position: Position?) : this() {
         if (position != null) {
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -1,18 +1,21 @@
 package com.strumenta.kolasu.model
 
-import com.strumenta.kolasu.mapping.position
-import org.antlr.v4.runtime.ParserRuleContext
+interface WithPosition {
+    val position: Position?
+}
+
+class JustPosition(override val position: Position): WithPosition
 
 /**
  * The Abstract Syntax Tree will be constituted by instances of Node.
  */
-open class Node(
-    /**
-     * Explicitly set position.
-     * @see Node.position
-     */
-    open var specifiedPosition: Position? = null
-) {
+open class Node(): WithPosition {
+
+    constructor(position: Position?) : this() {
+        if (position != null) {
+            origin = JustPosition(position)
+        }
+    }
 
     open val nodeType: String
         get() = this.javaClass.canonicalName
@@ -29,9 +32,10 @@ open class Node(
         }
 
     /**
-     * The parse-tree node from which this AST Node has been generated, if any.
+     * The node from which this AST Node has been generated, if any.
      */
-    var parseTreeNode: ParserRuleContext? = null
+    var origin: WithPosition? = null
+
     /**
      * The parent node, if any.
      */
@@ -42,8 +46,8 @@ open class Node(
      * If a position has been provided when creating this node, it is returned.
      * Otherwise, the value of this property is the position of the original parse tree node, if any.
      */
-    val position: Position?
-        get() = specifiedPosition ?: parseTreeNode?.position
+    override val position: Position?
+        get() = origin?.position
 }
 
 /**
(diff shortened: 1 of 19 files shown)


core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -1,18 +1,27 @@
 package com.strumenta.kolasu.model
 
-import com.strumenta.kolasu.mapping.position
-import org.antlr.v4.runtime.ParserRuleContext
+interface WithPosition {
+    val position: Position?
```
ftomassetti, line 4: Would it make sense to be able to change the position, for example to translate the node?
In the RPG parser what I do is parsing a subset of the text and then translating the nodes. Let's suppose I parse only lines 3, I would get nodes which have positions on line 1 (because I only passed one line to the parser). I would then translate them by two lines to get the correct positions

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -1,18 +1,27 @@
...
-) {
+open class Node() : WithPosition {
+
+    constructor(position: Position?) : this() {
+        if (position != null) {
+            origin = JustPosition(position)
+        }
+    }
+
+    constructor(origin: WithPosition?) : this() {
```
ftomassetti, line 20: Do all origins have a position? What if I create nodes arbitrarily? Would it be useful to have a type named Origin?

core/src/main/kotlin/com/strumenta/kolasu/parsing/ParseTree.kt
```diff
@@ -117,6 +116,44 @@ fun Token.getOriginalText(): String {
     return this.inputStream.getText(interval)
 }
 
-fun Node.getText(): String? = parseTreeNode?.getOriginalText()
+fun Node.getText(): String? = if (origin != null) {
```
ftomassetti, line 119: I think you could do something like `origin?.apply/let { }`

emf/src/test/kotlin/com/strumenta/kolasu/emf/rpgast/PredefinedIndicatorExpr.kt
```diff
@@ -3,7 +3,7 @@ package com.smeup.rpgparser.parsing.ast
 import com.strumenta.kolasu.model.Position
 
 // *IN01..*IN99
-data class PredefinedIndicatorExpr(val index: Int, override var specifiedPosition: Position? = null) :
+data class PredefinedIndicatorExpr(val index: Int, val specifiedPosition: Position? = null) :
```
ftomassetti, line 6: I would expect to be able to assign the specifiedPosition later on

ftomassetti (approved):
I think I would add a more generic type called Origin instead of WithPosition.
From my point of view, I would be ok with having a breaking change and move to 1.5, without the need to move to 2.0

ftomassetti:
Thank you for writing this!

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
alessiostalla: It's not mutable in the interface because some classes may compute it automatically. However, subclasses could have an updatePosition method or similar.

core/src/main/kotlin/com/strumenta/kolasu/parsing/ParseTree.kt
alessiostalla: Didn't know that, sure

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
ftomassetti: I think subclasses could make val into var, that should be permitted

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
ftomassetti: Would it make sense to have origin has something more generic, so that it can indicate not only origin through parsing but alternatively something like "GeneratedByProgramX" or "Obtained By transforming  AST node Y" ?

alessiostalla: In Origin, the position is optional, so yes indeed renaming WithPosition to Origin would be appropriate

core/src/main/kotlin/com/strumenta/kolasu/model/Printing.kt
```diff
@@ -18,9 +18,8 @@ fun Node.relevantMemberProperties(withPosition: Boolean = false, withNodeType: B
             !it.name.startsWith("component") &&
                 (it.name != "position" || withPosition) &&
                 (it.name != "nodeType" || withNodeType) &&
-                it.name != "specifiedPosition" &&
                 it.name != "properties" &&
-                it.name != "parseTreeNode" &&
+                it.name != "origin" &&
```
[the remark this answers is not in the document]
alessiostalla: Actually, it looks like DEFAULT_IGNORED_PROPERTIES is redundant and unnecessary since _nodeProperties_ are already filtered. But we cannot use _nodeProperies_ here because it excludes the position.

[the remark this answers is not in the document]
alessiostalla: And, actually, some properties are already filtered as _@Derived_, so probably another improvement would be to get rid of these lists of properties and just use _@Derived_ everywhere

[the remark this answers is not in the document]
ftomassetti: I would not use @Derived but probably creating a separate annotation as the meaning is a bit different. Derived here means "calculated from other values"

emf/src/test/kotlin/com/strumenta/kolasu/emf/rpgast/PredefinedIndicatorExpr.kt
ftomassetti: Good point!

core/src/main/kotlin/com/strumenta/kolasu/model/Printing.kt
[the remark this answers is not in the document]
alessiostalla: Ok, I thought of `@Derived` b/c _Node_ already has its _properties_ property marked as `@Derived`, but we could introduce another annotation, say `@BuiltIn`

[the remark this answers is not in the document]
alessiostalla: I've settled on `@Internal` for this.

alessiostalla:
@ftomassetti when all the tests pass, let's say this becomes kolasu v1.5.0-SNAPSHOT.
What happens to the current master branch? IMHO we should do a last 1.4.x release if appropriate. Then we could merge this into master. It would still make sense in the future to cut a v1.4.x branch so that we can fix bugs on 1.4.x versions without forcing existing projects to migrate in order to receive the fix.

PR #83, opened 2022-06-29
Track the "source" of each Position if provided

alessiostalla:


alessiostalla:
I've done this change first in Pylasu because I realized that even though we have the Position.source field we never set it and we don't offer a way to propagate it when computing the position of nodes. I had replicated a similar field to track issues over multiple files, but issues already have a position and that position already has a source. So I think we can spare users from manually tracking it, simplifying their design.

core/src/main/kotlin/com/strumenta/kolasu/model/Printing.kt
```diff
@@ -23,6 +23,7 @@ fun Node.relevantMemberProperties(withPosition: Boolean = false, withNodeType: B
                 it.name != "origin" &&
                 it.name != "destination" &&
                 it.name != "sourceText" &&
+                it.name != "source" &&
```
ftomassetti, line 26: This is a method that we should probably revise at some point to use the Internal annotation, but I am adding this just as a reminder, not to have this fixed as part of this PR

core/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -314,15 +314,7 @@ fun Node.processConsideringDirectParent(operation: (Node, Node?) -> Unit, parent
...
-        val children = mutableListOf<Node>()
-        this.properties.forEach { p ->
-            val v = p.value
-            when (v) {
-                is Node -> children.add(v)
-                is Collection<*> -> v.forEach { if (it is Node) children.add(it) }
-            }
-        }
-        return children
+        return walkChildren().toList()
```
ftomassetti, line 317: Better :D

core/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
```diff
@@ -335,8 +327,7 @@ fun Node.transformTree(
     mutationsCache.computeIfAbsent(this) { operation(this) }
     val changes = mutableMapOf<String, Any>()
     relevantMemberProperties().forEach { p ->
-        val v = p.get(this)
-        when (v) {
+        when (val v = p.get(this)) {
```
ftomassetti, line 330: nice, I did not know you could do that!

ftomassetti:
+1 Good work. Keeping the *Lasu aligned is a very good idea

ftomassetti (approved):
+1

PR #85, opened 2022-07-08
Introduce SyntheticSource

### LINKED ISSUE #60: Marker for synthetic/generated nodes
Some nodes are calculated, for example the CharacterType node in the RPG parser. We could mark them appropriately to indicate they have a reason for not having a position

ftomassetti:
This should fix #60


ftomassetti:
@alessiostalla do you think this make sense?

alessiostalla:
Yes I think it makes perfect sense!

ftomassetti:
Thank you, I will merge it then

Issue #78, opened 2022-06-06
Revise support for CLI utilities


ftomassetti:
Expand it to include code from RPGParser

ftomassetti:
See:
https://github.com/Strumenta/kolasu/blob/master/core/src/main/kotlin/com/strumenta/kolasu/cli/CLI.kt
https://github.com/Strumenta/kolasu/blob/master/emf/src/main/kotlin/com/strumenta/kolasu/emf/cli/ParserCLIEMF.kt

ftomassetti:
In core we may want to:
* Generate ASTs on file (as XML or JSON)
* Produce statistics on the number of errors
* Produce statistics on the types of nodes

We may want to specify the character encoding and verbosity

ftomassetti:
I am reflecting on which command-line utilities we need.

I find it useful to have command-line tools to:
- Export the AST (XML, JSON, Debug format) on file or the command line
- Export the metamodel in EMF format (XMI, JSON)
- Export ASTs in EMF format (XMI, JSON)
- Print stats on the presence of nodes in a set of source files. @mctaverna and @loradd worked on similar tools for some analysis we did. What are your thoughts on this?
- Print stats on the errors that occurred while parsing files

@alessiostalla wrote two command line tools (one in core and one in EMF) which performed most of these things. Any suggestion for this?
I have been doing some experiments in [branch issue78](https://github.com/Strumenta/kolasu/tree/issue78).
So far, I have focused only on core, and I wanted to create two separate commands: one for stats and one for AST exporting.

alessiostalla:
I don't have any suggestions...

mctaverna:
@ftomassetti 
In my opinion, it would be useful to print statistics about the number of nodes found in a single or in a set of files. 
Something like SELECT DISTINCT node, count(*) GROUP BY node for all nodes in the AST.
The report should include the nodes not present (count = 0) in the file/file set. 
It would be helpful to filter the result by excluding for example the expression subclasses (i.e.Literals) because are not very interesting.

ftomassetti:
Closed by #88

PR #88, opened 2022-07-14
Revising CLI Tool support

ftomassetti:
We aim to solve #78 

The idea is to revise CLI tools, to avoid having to reimplement the same logic in every project. Over time we could build more features and options, making these CLI tools more powerful.

My main concern is regarding EMF support: I tried reusing the code written by @alessiostalla, but that part is not clear to me, so the opinion of @alessiostalla would be particularly useful.


core/src/main/kotlin/com/strumenta/kolasu/cli/ASTProcessingCommand.kt
```diff
@@ -0,0 +1,134 @@
...
+import com.strumenta.kolasu.validation.IssueSeverity
+import com.strumenta.kolasu.validation.Result
+import java.io.File
+import java.nio.charset.Charset
+import java.util.function.Function
+import kotlin.system.exitProcess
+
+typealias ParserInstantiator<P> = Function<File, P?>
+
+abstract class ASTProcessingCommand<R : Node, P : ASTParser<R>>(
```
ftomassetti, line 23: The idea is to have this base class for all commands that could process entire directories containing source code and produce a sequence of ASTs, each to be processed

core/src/main/kotlin/com/strumenta/kolasu/cli/CLI.kt
```diff
@@ -1,90 +0,0 @@
...
-    allowMultipleSubcommands: Boolean = false,
-    treatUnknownOptionsAsArgs: Boolean = false,
-    val exitOnFail: Boolean = true,
-    val printASTSummary: Boolean = true
-) : CliktCommand(
-    help, epilog, name, invokeWithoutSubcommand, printHelpOnEmptyArgs, helpTags, autoCompleteEnvvar,
-    allowMultipleSubcommands, treatUnknownOptionsAsArgs
-) {
-    var result: ParsingResult<R>? = null
-    val context by requireObject<CLIContext<R>>()
```
ftomassetti, line 44: I do not understand what context is used for

core/src/main/kotlin/com/strumenta/kolasu/cli/CLI.kt
```diff
@@ -1,90 +0,0 @@
...
-                "(${parseTree.start.inputStream.size()} tokens)."
-        )
-        if (printASTSummary) {
-            println("The AST has ${countASTNodes(node)} nodes.")
-        }
-    }
-
-    fun countASTNodes(node: Node): Int = 1 + (node.children.map { countASTNodes(it) }.reduceOrNull(Int::plus) ?: 0)
-
-    fun countParseTreeNodes(parseTree: ParseTree): Int {
```
ftomassetti, line 83: This part has been removed as in general I would tend to hide parse trees under the hood: we may not always have them. Perhaps instead I am wrong and these things are useful while developing parsers based on ANTLR

emf/src/main/kotlin/com/strumenta/kolasu/emf/cli/EMFCLICommands.kt
```diff
@@ -0,0 +1,106 @@
...
+        .flag(default = false)
+    val output by option("--output", "-o")
+        .file()
+        .help("File where to store the metamodel")
+        .default(File("metamodel.json"))
+    val includeKolasu by option("--include-kolasu", "-ik").flag(default = false)
+
+    override fun run() {
+        val metamodel = metamodelGenerator.get()
+        val mmResource = JsonResourceFactory().createResource(URI.createFileURI(output.path))
```
alessiostalla, line 97: The original ParserCLIEMF used the resource returned from `EMFMetamodelSupport.saveMetamodel` whose `ResourceFactory` is initialized like this:
```
resourceSet.resourceFactoryRegistry.extensionToFactoryMap["json"] = JsonResourceFactory()
    resourceSet.resourceFactoryRegistry.extensionToFactoryMap["xmi"] = XMIResourceFactoryImpl()
    resourceSet.resourceFactoryRegistry.extensionToFactoryMap["ecore"] = EcoreResourceFactoryImpl()
```

So it can save to .json, .xmi and .ecore. I think we should preserve at least .json and .xmi, not sure about .ecore (maybe @loradd can comment on the difference between the two? I seem to recall that .ecore only supports one EPackage and silently drops the others).

emf/src/main/kotlin/com/strumenta/kolasu/emf/cli/EMFCLICommands.kt
```diff
@@ -0,0 +1,106 @@
...
+        }
+        val mmResource = parser.saveMetamodel(URI.createFileURI(target.path))
+        if (verbose) {
+            echo("Done (${System.currentTimeMillis() - start}ms).")
+        }
+        return mmResource
+    }
+}
+
+class EMFMetaModelCommand(val metamodelGenerator: Supplier<EPackage>) :
```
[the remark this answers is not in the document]
ftomassetti: Right, I did not consider the case of multiple packages.
I initially did not consider EMFEnabledParser because I thought some parsers required an input to be instantiated, but I now realize I am wrong. So these two options you specified should work.  I will go for EMFMetamodelSupport

emf/src/main/kotlin/com/strumenta/kolasu/emf/cli/EMFCLICommands.kt
ftomassetti: Good point. I created utility methods for creating resourceSets and resources and I tried to use them consistently in the project

ftomassetti:
I applied the suggestions and added tests

emf/src/main/kotlin/com/strumenta/kolasu/emf/cli/EMFCLICommands.kt
loradd: > (maybe @loradd can comment on the difference between the two? I seem to recall that .ecore only supports one EPackage and silently drops the others).

Not sure about this, I always used `XMIResourceFactory`.

Looking at the code, both factories seem to create resources extending `XMLResource`.  `EcoreResourceFactory` seems to configure details about IDs, URIs and encoding, while `XMIResource` is an extension of `XMLResource` in `XMIResourceFactory`.

```java
// EcoreResourceFactoryImpl
@Override
public Resource createResource(URI uri) {
    XMLResource result = new XMLResourceImpl(uri) {
        @Override
         protected boolean useIDs() {
             return eObjectToIDMap != null || idToEObjectMap != null;
         }
     };
     result.setEncoding("UTF-8");
     result.getDefaultSaveOptions().put(XMLResource.OPTION_USE_ENCODED_ATTRIBUTE_STYLE, Boolean.TRUE);
     result.getDefaultSaveOptions().put(XMLResource.OPTION_LINE_WIDTH, 80);
     result.getDefaultSaveOptions().put(XMLResource.OPTION_URI_HANDLER, new URIHandler.PlatformSchemeAware());
        return result;
}

// XMIResourceFactoryImpl
@Override
public Resource createResource(URI uri) {
    return new XMIResourceImpl(uri);
}
```

PR #96, opened 2022-08-02
Fix how generics are handled in EMF metamodel serialization (Kolasu 1.4)

ftomassetti:
@loradd reviewing this is NOT urgent

### DIFF
--- a/emf/src/main/kotlin/com/strumenta/kolasu/emf/Metamodel.kt
+++ b/emf/src/main/kotlin/com/strumenta/kolasu/emf/Metamodel.kt
@@ -13,10 +13,7 @@ import java.time.LocalDateTime
 import java.time.LocalTime
 import java.util.*
 import kotlin.collections.HashMap
-import kotlin.reflect.KClass
-import kotlin.reflect.KClassifier
-import kotlin.reflect.KType
-import kotlin.reflect.KTypeParameter
+import kotlin.reflect.*
 import kotlin.reflect.full.*
 
 interface EDataTypeHandler {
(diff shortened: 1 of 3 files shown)


ftomassetti:
If and when this get merged, we should then port this fix to the master branch (wip in branch `forwardPortGenericFix`)

emf/src/main/kotlin/com/strumenta/kolasu/emf/Metamodel.kt
```diff
@@ -290,33 +216,29 @@ class MetamodelBuilder(packageName: String, nsURI: String, nsPrefix: String, res
         eClass.isAbstract = kClass.isAbstract || kClass.isSealed
         eClass.isInterface = kClass.java.isInterface
 
+        kClass.typeParameters.forEach { kTypeParameter: KTypeParameter ->
+            eClass.eTypeParameters.add(
+                EcoreFactory.eINSTANCE.createETypeParameter().apply {
+                    name = kTypeParameter.name
+                }
+            )
+        }
```
loradd, line 225: This looks good, just a comment.

We do not support bounds, e.g. A<B :  C>, and more complex specifications involving generics, e.g. A<D, B: C<D>>. I think we should keep track of this, in case it turns out to be required (not sure). 

Ideally, this could be achieved processing `kTypeParamater.upperBounds` and map these into `EGenericType` instances (similarly to those created for references).

loradd (commented):
It looks good to me, I left a comment for a possible future extension covering more complex generic type specifications (not sure it is going to be needed).

emf/src/main/kotlin/com/strumenta/kolasu/emf/Metamodel.kt
ftomassetti: Right, we should add a comment. For implementing support however we should consider that  Kotlin has this concept of  out and in variant which I do not think map exactly to things in EMF, so I am not sure we can support them properly.

loradd: I agree, maybe the variance information could be annotated using an EAnnotation.

ftomassetti: Good idea!

PR #102, opened 2022-08-26
Added indexing support for computing node ids + ReferenceByName support in JSON serialization

### LINKED ISSUE #101: Handle ReferenceByName in native serialization
The current native JSON serialization should be modified to handle ReferenceByName with dedicated logic. Given a resolved ReferenceByName, we should avoid re-serializing the target node and use ids or similar link approaches.

loradd:
Closes #101 

Signed-off-by: Lorenzo Addazi <lorenzo.addazi@strumenta.com>


core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * @param walker a function that generates a sequence of nodes. By default this is the depth-first "walk" method.
+ * @param idProvider a function that takes a node and returns a String identifier representing it
+ * For post-order traversal, take "walkLeavesFirst".
+ * @return walks the whole AST starting from the given node and associates each visited node with a generated id
+ **/
+fun Node.computeIds(
+    walker: (Node) -> Sequence<Node> = Node::walk,
+    idProvider: (Node) -> String = { UUID.randomUUID().toString() }
```
ftomassetti, line 14: For the default implementation I would consider a mechanism that provides sequential IDs, so that they are shorter. The advantage of shorter IDs is that the resulting files are smaller and they are more "human friendly"

core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
```diff
@@ -136,6 +164,17 @@ class JsonGenerator {
         }
     }
 
+    private fun computeIds(root: Node): IdentityHashMap<Node, String> =
+        IdentityHashMap<Node, String>().apply {
+            root.walk().forEach { this[it] = UUID.randomUUID().toString() }
```
ftomassetti, line 169: I would consider assigning IDs only to nodes which needs them as someone is referring to them. In this way it will make easier to find the referred nodes

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
loradd: Do you mean using something similar to a counter?

ftomassetti: Nice touch to have an idProvider!

ftomassetti (approved):
I have some suggestions for performance and readability reasons, but they are secondary and in general the solution seems good

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
ftomassetti: Yes, a counter would be perfect

core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
loradd: In order to be able to do this, `ReferenceByName` should be modelled as `NodeType` as the current walkers cannot retrieve these otherwise. Am I correct?

loradd: Another possibility would be to process the properties, as follows:
```
val ids = node.computeIds {
    it.walk()
        .find { node -> node.properties
            .any { property -> property.value is ReferenceByName<*>
                     && (property.value as ReferenceByName<*>).referred == it}}  
        ?.let { "some_identifier" }
}
```

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
loradd: I have defined an `IdProvider` interface with a `SequentialIdProvider` default implementation for such purpose. Now sequential integers are used in place of random UUIDs. Custom IdProviders can be implemented as lambda, if no local state is required, like so:
```
node.computeIds {
    node -> some_function_returning_string(node) // this will be the node id
}
```
Otherwise, stateful idProviders can be implemented as the default `SequentialIdProvider` or `OnlyReferencedIdProvider`, like:
```
class SequentialIdProvider(private var counter: Int = 0) : IdProvider {
    override fun getId(node: Node): String? {
        return "${this.counter++}"
    }    
}
```

`OnlyReferencedIdProvider` is used to limit id computation to nodes that have at least one ReferenceByName targeting them.

core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
loradd: `OnlyReferencedIdProvider` serves this purpose, it internally uses `SequentialIdProvider` by default but can be easily paremeterized.

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
```diff
@@ -12,6 +12,16 @@ class SequentialIdProvider(private var counter: Int = 0) : IdProvider {
     }
 }
 
+class OnlyReferencedIdProvider(
+    private var idProvider: IdProvider = SequentialIdProvider()
+) : IdProvider {
+    override fun getId(node: Node): String? {
+        return node.properties
```
ftomassetti, line 19: Doesn't this only select nodes that refer to themselves?

ftomassetti: I think the work to understand which nodes are referred to should be done by who creates the value for withIds before calling generateJson. They can do that traversing the whole AST

loradd: > Doesn't this only select nodes that refer to themselves?

You're right, my mistake, should be fixed now.

> I think the work to understand which nodes are referred to should be done by who creates the value for withIds before calling generateJson. They can do that traversing the whole AST

Done. The `generateJson` call now invokes `computeIdsForReferencedNodes`, which computes ids using the (no more default) ReferencedOnlyIdProvider id provider.

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
```diff
@@ -0,0 +1,49 @@
...
+        return "${this.counter++}"
+    }
+}
+
+class OnlyReferencedIdProvider(
+    private var idProvider: IdProvider = SequentialIdProvider()
+) : IdProvider {
+    override fun getId(node: Node): String? {
+        val root = (node.walkAncestors().lastOrNull() ?: node).apply { assignParents() }
+        return root.walk().find { otherNode ->
```
ftomassetti, line 20: We traverse the whole AST at each request. I think we should either pre-calculating which nodes are referred to at construction time, or cache results

core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
ftomassetti: I think the alternative approach (which you implement) is better as the ReferenceByName does not represent a node per se, but a reference to a node

core/src/test/kotlin/com/strumenta/kolasu/model/IndexingTest.kt
```diff
@@ -0,0 +1,80 @@
+package com.strumenta.kolasu.model
+
+import org.junit.Test
+import kotlin.test.assertContains
+import kotlin.test.assertEquals
+
+class IndexingTest {
+
+    @Test
+    fun computeIdsWithDefaultWalkerAndIdProvider() {
```
ftomassetti, line 10: I think it would be useful to have a test showing the IdProvider that assigns IDs only to Node which are referred. When used in this case, I think that b1 should not have an ID

loradd: This should be covered now

core/src/main/kotlin/com/strumenta/kolasu/model/Indexing.kt
loradd: This should be covered now

Issue #97, opened 2022-08-03
Transpilation trace: track issues


alessiostalla:
We should add an issues property, akin to issues in ParsingResult

PR #106, opened 2022-09-01
Add issues to the transpilation trace

ftomassetti:
We add both issues for the source and target AST but also for the transpilation itself. For that we also introduce a new IssueType.

This solves #97 . Adding @alessiostalla as a reviewer since he opened that issue


alessiostalla:
I understand sourceResult but what kind of issues would go into targetResult as compared with the top-level issues collection?

In the transpiler I'm working on, transpilation issues are represented as SEMANTIC issues on the target result. I suppose those would be the only kind of issues you can possibly find on the target, though I appreciate that we can distinguish them with a new issue type (maybe I'd call it TRANSLATION or GENERATION which is a bit more general than TRANSPILATION).

ftomassetti:
Good question. Perhaps we could analyze the result and produce errors. For example, an unsolved symbol in the generated code, or generated code that is invalid. The issues on the top-level collection would refer instead to errors in the generation process itself.

+1 for renaming in translation

ftomassetti:
Visually the difference between the kind of issues would be that issues on the target would be displayed in the generated code panel, while issues for the transpilation would appear separately

alessiostalla:
Makes sense!

PR #107, opened 2022-09-02
Making the core module independent from ANTLR

ftomassetti:
In general, not all ASTs we may need to build have to be obtained from translating an ANTLR parse tree.
We could potentially obtain the AST from deserialization, from other parser generators or parsing approaches (for example, when working with column based languages or binary formats). Or in the future we may have have nodes that obtained from textual representations.

For these reasons we extract ANTLR into a separate module and make the core Kolasu module independent.

One challenge is that the EMF module would need ANTLR to define the EMFEnabledParser class in a way that it extends the KolasuANTLRBasedParser (provided by the antlr module). If EMFEnabledParser does not extends KolasuANTLRBasedParser, it would make complicated to create parsers both based on ANTLR and needing EMF support. We then removed the EMFENabledParser and just use the EMFMetamodelSupport interface. This would cause a little bit more work for the implementation of parsers extending KolasuANTLRBasedParser and implementing EMFMetamodelSupport (basically copying a 4 lines method)


ftomassetti:
In ParsingResult I remove the firstStage property, as not all parser may go through a first stage.
I instead introduce this:

val intermediateResults: Map<String, Any>

I am not sure if this is a good idea or we should instead have an ANTLRParsingResult extending a more generic ParsingResult.

ftomassetti:
This should consider #108

ftomassetti:
Closing: we could consider this later, as part of 1.6

PR #118, opened 2022-09-06
Updates to Origin and Destination

ftomassetti:
This PR is based on discussion https://github.com/Strumenta/StarLasu/discussions/5

This avoids touching `positionOverride` but focus only on the other, less controversial changes.

It includes update to EMF support


alessiostalla (approved):
Looks good to me!

PR #124, opened 2022-09-09
#123 Use native Ecore EBigInteger/Decimal instead of ad-hoc custom types

alessiostalla:


ftomassetti:
+1, thank you for correcting this one!

PR #129, opened 2022-09-21
Issue 128: Serialization of transpilation traces

ftomassetti:
I started adding a test for this. @alessiostalla was your test different? Should we add also that one?


alessiostalla:
No it's the same, sorry I was convinced I had pushed it

ftomassetti:
No problem!

Issue #131, opened 2022-09-22
Serialization of AST's with ParseTreeOrigin doesn't work


alessiostalla:

ftomassetti:
This is serialization to EMF-JSON, right? Not to native JSON or XML

I am not sure if this could work: I mean, should we aiming to replicate the whole parse-tree? On the other side they could not have the generated parse-tree classes to unserialize it anyway

alessiostalla:
No, my idea is to ignore the parse tree origin and only save the position.
In any case, this should work (i.e. don't throw an exception) out of the
box.

Il ven 23 set 2022, 08:49 Federico Tomassetti ***@***.***> ha
scritto:

PR #132, opened 2022-09-29
Issue 131: serialization to EMF of ParseTreeOrigin

ftomassetti:
To support this we needed to transform the ASTNode.origin link from a reference link to a containment link. We therefore needed to introduce NodeOrigin as a wrapper to a reference to Node


ftomassetti:
Aims to fix #131

alessiostalla:
I thought we could just leave origin = null and set the position of the node when the origin is a ParseTreeOrigin. Right now, ParseTreeOrigin in the Ecore export only repeats the position of the node, so it looks kind of redundant. At least we should omit the position in ParseTreeOrigin, I think.
Anyway, I think changing origin from reference to containment is the right choice to support other kinds of non-Node origins in the future.

ftomassetti:
I modified it so that ParseTreeOrigin is not saved but the position is

PR #142, opened 2022-10-17
Initial implementation of the code generation support

ftomassetti:
This code has been used so far in the StarLasu-tools to generate Kotlin code, in the transpiler demo being prepared for Codemotion to generate Java code, by @loradd to generate Python code.

It may be useful to have a preliminary version in a release version of Kolasu in order to use it more, validate it and possibly improve it.

We would need to understand if this feature makes the `Unparser` class redundant or not.

It would be useful to have the opinion of @alessiostalla as this touches the Unparser he wrote.
It would be useful to have the opinion of @loradd as he as used these APIs in the Python Code generator.


alessiostalla (approved):
Looks good!

ftomassetti:
Thank you @alessiostalla !

loradd (approved):
Looks good! 

We might want to move the Kotlin AST out somewhere in the future.

ftomassetti:
Thank you @loradd !

Regarding the Kotlin AST, this is just a toy version for tests. We have a better one in StarLasu-tools and I agree that we should move that one into a separate project

PR #148, opened 2022-10-24
ASTTransformer: post processing behavior

### LINKED ISSUE #147: ASTTransformer: post processing behavior
When transforming ASTs we want to add an hook to execute some code after having transformed all children

loradd:
Closes #147 

Signed-off-by: Lorenzo Addazi <lorenzo.addazi@strumenta.com>

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
@@ -22,18 +22,9 @@ annotation class Mapped(val path: String = "")
  */
 class NodeFactory<Source, Output : Node>(
     val constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> Output,
-    val children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf()
+    val children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf(),
+    var finalizer: (Output) -> Unit = {}
 ) {
-
-    private var finalizer: (Output) -> Unit = {}
-
-    fun finally(node: Output) {
-        this.finalizer(node)
-    }
-
-    fun finally(finalizer: (Output) -> Unit) {
-        this.finalizer = finalizer
-    }
 
     fun withChild(
         sourceProperty: KProperty1<Source, *>,
@@ -69,6 +60,10 @@ class NodeFactory<Source, Output : Node>(
         val prefix = if (type != null) type.qualifiedName + "#" else ""
         children[prefix + name] = ChildNodeFactory(prefix + name, get, set)
         return this
+    }
+
+    fun withFinalizer(finalizer: (Output) -> Unit) {
+        this.finalizer = finalizer
     }
 
     fun getter(path: String) = { src: Source ->
@@ -183,7 +178,7 @@ open class ASTTransformer(
                     }
                 }
             }
-            factory.finally(node)
+            factory.finalizer(node)
             node.parent = parent
         } else {
             if (allowGenericNode) {
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
@@ -3,11 +3,14 @@ package com.strumenta.kolasu.transformation
 import com.strumenta.kolasu.model.*
 import com.strumenta.kolasu.validation.Issue
 import com.strumenta.kolasu.validation.IssueSeverity
-import java.lang.IllegalStateException
 import kotlin.reflect.KClass
 import kotlin.reflect.KMutableProperty1
 import kotlin.reflect.KProperty1
-import kotlin.reflect.full.*
+import kotlin.reflect.full.createInstance
+import kotlin.reflect.full.findAnnotation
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -17,15 +20,26 @@ annotation class Mapped(val path: String = "")
 /**
  * Factory that, given a tree node, will instantiate the corresponding transformed node.
  */
-class NodeFactory<Source>(
-    val constructor: (Source, ASTTransformer, NodeFactory<Source>) -> Node,
+class NodeFactory<Source, Output : Node>(
```
ftomassetti, line 23: I verified that this additional type parameter does not seem to require an extra declaration in the examples I have developed for the demo (the system is still able to infer the value), so that looks good to me

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -17,15 +20,26 @@ annotation class Mapped(val path: String = "")
...
+    val constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> Output,
     val children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf()
 ) {
+
+    private var finalizer: (Output) -> Unit = {}
+
+    fun finally(node: Output) {
```
ftomassetti, line 30: If I understand correctly this invoke the finalizer, while the following method with the same name set it. That could be maybe confusing

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -17,15 +20,26 @@ annotation class Mapped(val path: String = "")
...
+        this.finalizer(node)
+    }
+
+    fun finally(finalizer: (Output) -> Unit) {
```
ftomassetti, line 34: Should this be a property?

core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
```diff
@@ -116,4 +137,110 @@ class ASTTransformerTest {
             )!!
         )
     }
+
+    /**
+     * Example of transformation to perform a simple type calculation.
+     */
+    @Test
+    fun computeTypes() {
+        val myTransformer = ASTTransformer(allowGenericNode = false).apply {
```
ftomassetti, line 146: Would it make sense to later introduce a specific subtype of ASTTransformer specifically for type calculations?

ftomassetti (approved):
It looks good to me. I added a few minor comments. Please feel free to take a look at them and then merge directly this PR, just let me know if you need permissions to be assigned to you in order to do the merge

core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
loradd: It could make sense, but I think we could also investigate if having type computation and symbol resolution implemented as lazy properties of the AST nodes would come along better.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
loradd: The reason why I splitted `finally` and `finalizer` is to allow users to define the `finalizer` in a more fluid way. That is:
```kotlin
registerNodeFactory(...)
    .withChild(...)
    .finally { ... }
```
rather than
```kotlin
registerNodeFactory(...)
    .withChild(...)
    .apply { this.finalizer = {...} }
```
or 
```
val nodeFactory = registerNodeFactory(...).withChild(...)
nodeFactory.finalizer = {...}
```
Does it make sense? Maybe a more explicit name like `onAllChildrenProcessed` could work better than `finalizer` and `finally` though.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
loradd: See discussion below.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
loradd: The additional type parameter has been introduced to avoid additional casts in 
```kotlin
finally { targetNode -> ...targetNode as SomeNodeSubClass... }
```
but we could remove it in the future if we see that the change causes issues.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
ftomassetti: We discussed this in a meeting and Lorenzo suggested renaming this one `finalize`

alessiostalla:
What about finalizeWith or withFinalizer to keep  the "with" convention?

Il mar 25 ott 2022, 09:24 Federico Tomassetti ***@***.***> ha
scritto:

loradd:
As @alessiostalla suggested, I renamed `finally` into `withFinalizer` to be more consistent with the `withChild` methods. The property `finalizer` has been made public, similarly to `constructor`.

PR #149, opened 2022-10-25
Feature: node detach method

alessiostalla:


core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -13,6 +13,8 @@ interface Origin {
         get() = position?.source
 }
 
+class DetachedOrigin(override val position: Position?, override val sourceText: String?) : Origin
```
ftomassetti, line 16: Could this be useful also when describing the origin without deriving it from a ParseTreeNode?
For example, building it manually by setting position and sourceText explicitly?
In that case, could it make sense to rename it?

ftomassetti (approved):

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
alessiostalla: I agree, it could be useful that way too. We could name it `UnspecifiedOrigin` or maybe `SimpleOrigin`, do you have any better ideas for a name?

PR #152, opened 2022-11-23
Refining of the AST Transformations

ftomassetti:
Initial draft, to get some feedback.

I am working on implementing Parse Tree to AST transformations in the easiest way possible, considering that in the AST I would like to use constructors with parameters.

In the ideal scenario we can just do this:

```
transformer.registerTrivialPTtoASTConversion<ScriptContext, SScript>()
```

In this case the system check that SScript expects one constructor parameter and that is populated with the corresponding value (same name) in the parse tree.

In some cases we change some names so we may want to specify that:

```
transformer.registerTrivialPTtoASTConversion<Int_literal_expressionContext, SIntegerLiteral>(
            Int_literal_expressionContext::INT_VALUE to SIntegerLiteral::value
)
```

I am not sure how to set the parent correctly and I would check with the author of the ASTTransformation approach if this makes sense


alessiostalla:
This looks interesting. To set the parent of constructor arguments you could save them in a list and after having invoked the constructor you can set the parent on the saved children. Or just iterate on the new node's children and set the parent.

I'm wondering if the logic to call the constructor and to automatically translate strings etc. could be made part of the transformer itself. For sure I don't like having to use `var` just to make the transformer happy. But I think we should leave that to future issues.

ftomassetti:
> This looks interesting. To set the parent of constructor arguments you could save them in a list and after having invoked the constructor you can set the parent on the saved children.

Right, why did I not think about that!

ftomassetti:
> I'm wondering if the logic to call the constructor and to automatically translate strings etc. could be made part of the transformer itself. For sure I don't like having to use var just to make the transformer happy. But I think we should leave that to future issues.

I would be ok doing so, but perhaps we can check how well that works this logic (which is now in a separate class) and if that works well enough we could move it into the ASTTransformer class with 1.6.0. Would this makes sense?

alessiostalla:
Yes, that makes a lot of sense.

ftomassetti:
Ok, then I think this is reading for your review. Given it adds to ASTTransformer but does not change it at this stage, I think the bar is lower

PR #153, opened 2022-12-09
Implementing Node.toString

ftomassetti:
This implementation avoid the risk of stack overflow in case of nodes using links and presenting circular graphs.

The drawback is that the implementation must be final, otherwise the auto-generated toString of data classes prevail.


alessiostalla (approved):

ftomassetti:
Thank you Alessio!

PR #157, opened 2023-01-09
Node printer overrider

ftomassetti:
Add a mechanism to override the behavior of an existing ASTCodeGenerator.
This is used for example to print special strings for node representing placeholders


alessiostalla (approved):
Looks good

ftomassetti:
Thanks!

PR #158, opened 2023-01-09
Quoted element refinements

ftomassetti:
Introducing QuotedElements. QuotedElement is used to represent elements in code templates and code matcher.
For example, in a code matcher: \`expr:v\` + 1, where \`expr:v\` is a QuotedElement, should match any code containing any expression added to 1.

In a code template \`expr:v\` + 1 indicates that, given expression `v` we want to sum it to 1.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/CommonElements.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/CommonElements.kt
@@ -12,6 +12,21 @@ interface Expression
 @NodeType
 interface EntityDeclaration
 
+/**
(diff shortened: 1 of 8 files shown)


alessiostalla:
My only objection would be the name "quoted" as quotation usually means to prevent evaluation, so that e.g. `1 + 2` => 3 but `quote(1 + 2)` => Sum(Lit(1), Lit(2)). I would rather use "PlaceholderNode" or something like that.

ftomassetti:
I have renamed `QuotedElement` to `PlaceholderElement` as suggested. Thank you!

Issue #160, opened 2023-01-10
Transformers: allow to opt out of setting children


alessiostalla:
Our SAS parser likes to configure some transformers like the following:

```kotlin
transformer.registerNodeFactory(XYZContext::class) { ctx -> transformer.transform(ctx.children[0]) }
```

E.g. when you have a context that is often just a wrapper over several alternatives and for some reason those are not labeled alternatives in ANTLR (subclasses).

With the original transformers implementation, somehow the above worked (even though it had the unwanted side-effect of going over the node's children twice). But with the Kolasu-native transformers, the above doesn't work when you use `@Mapped`, because the transformer gets a source tree that doesn't match with the transformed tree. 

Ideally, when a node factory already calls `transform`, there's no reason to go over the node's children again. So, we should have a way to tell the transformer that a given factory already takes care of the node's children and no further computation is desired on that subtree.

alessiostalla:
To be clearer, the issue with `@Mapped` is as follows.

Transformers require explicit mappings for children, in the form of "child node factories" that can be either configured with API calls, or declaratively with `@Mapped`. When using `@Mapped`, the transformer's configuration is determined dynamically using reflection. So, in a case like the following:

```kotlin
transformer.registerNodeFactory(XYZContext::class) { ctx -> transformer.transform(ctx.children[0]) }
```

If the result of `transformer.transform(ctx.children[0])` is an instance of a `Node` with a child annotated with `@Mapped("someProperty")`, the transformer will think that it has to populate that child, according to the configuration determined by reflection. However, the "source" of the node will be an instance of `XYZContext` that **does not** have a child named `someProperty`. So the transformer will fail.

PR #161, opened 2023-01-10
Transformers: allow to opt out of setting children 

alessiostalla:
Addresses issue #160

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
@@ -68,6 +68,21 @@ class NodeFactory<Source, Output : Node>(
         return this
     }
 
+    /**
+     * Tells the transformer whether this factory already takes care of the node's children and no further computation
+     * is desired on that subtree. E.g., when we're mapping an ANTLR parse tree, and we have a context that is only a
+     * wrapper over several alternatives, and for some reason those are not labeled alternatives in ANTLR (subclasses),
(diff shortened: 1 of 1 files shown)


ftomassetti:
I have two comments:
1) I think a comment about why this is useful could help. The reason is that the use-case for nodes that wrap other nodes (explained in the connected issue) is not that well-known, I think
2) I tend to think about all the nodeProperties have playing different roles: some contains children while other contains attributes. So if I read "skipChildren" I would expect that we only skip the properties that deal with nodes but we still consider the other properties (the one setting attributes: strings, ints, booleans, enums, etc.)

alessiostalla:
Regarding point 1 this is useful whenever the "node factory" already takes care of populating a node's children. This is the case of the SAS parser as explained in the issue, but also potentially when using "identity transformers". However, the issue only appears when one uses the `@Mapped` annotation, that's why it's not obvious. I've added more details to issue #160. Long-term, I think `@Mapped` should be deprecated and removed; however, the issue still remains **if** we want to continue supporting transforming from nested objects (e.g. child "a" of the transformed node is the transformation of child "b" of child "c" of the source node).

Regarding 2 transformers only deal with children, not with properties. Properties have to be set by the "Node factory" either manually or by reflection, but they're out of the transformer's scope.

ftomassetti:
Regarding point 1, sorry, I meant adding a comment in the code, so that a user of the library would know when to use that flag

Regarding point 2, ok, understood

ftomassetti:
+1, thanks Alessio

ftomassetti (approved):

PR #162, opened 2023-01-23
Fix method not found when of type list

jduchateau:
ANTLR is generating two members for children of list types, one returning the whole list, the other just one child when provided an index.
We always need the full list with a single parameter, otherwise we get `Exception java.lang.IllegalArgumentException: Callable expects 2 arguments, but 1 were provided` when using `registerTrivialPTtoASTConversion`.


ftomassetti:
hi @jduchateau , thank you for catching this issue and solving it!

PR #164, opened 2023-01-30
Improving performance of traversing AST

gabriele-tomassetti:
@ftomassetti can you confirm that the feature was implemented as we discussed?

We are trying to improve performance of traversing the AST. The current approach is to freeze the AST and record current children for each node.

Results seems promising when walking the frozen AST, but to freeze the AST it takes twice the time of a standard walk. Typical results:

```
Nodes: 702133
Walk time: 2648
Walk time frozen: 174
Freezing time: 5333
```

@alessiostalla does it seems a good approach to you?


alessiostalla:
Maybe instead of `if (map.has(key)) map.put(key, value)` you could try with `putIfAbsent`, I think in Java at least there's an overload of that method taking a closure, I don't know if Kotlin maps support a similar API.

alessiostalla:
At a higher level, what problem is this trying to solve? Is it because `children` calls `toList`? We could have a version of the children property that returns a sequence/generator (however it's called in Kotlin) to avoid constructing short-lived lists. Or is the use of reflection the main issue?

gabriele-tomassetti:
There is `putIfAbsent` method, I am going to update the code to use that.

I think that the main issue is the second one you mentioned: the use of reflection takes time. The `walk` method already uses a sequence.

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
```diff
@@ -159,15 +159,12 @@ fun <T> Node.collectByType(klass: Class<T>, walker: KFunction1<Node, Sequence<No
 
 fun Node.freeze() {
     walkLeavesFirst().forEach {
-        if(!Node.Freezer.childrenMap.containsKey(it)) {
-            Node.Freezer.childrenMap.put(it, it.children)
-        }
+        Node.childrenMap.putIfAbsent(it, it.children)
```
alessiostalla, line 162: This will still compute `it.children` every time though. I think there's an overload taking a closure. Anyway, probably `children` will be called only once so I don't think it will be a problem most of the time.

alessiostalla:
> I think that the main issue is the second one you mentioned: the use of reflection takes time. The `walk` method already uses a sequence.

Then, I don't think freezing buys you much here. Reflection has its own caches so probably if you just walk twice over the same tree the second time will be faster.
I think @ftomassetti has a tool to compute the reflection information ahead of time and patching the bytecode, so that it won't use reflection at run time.

ftomassetti:
The problem reported by the client is that traversing is slow. I think that we can often be in the situation in which a client has to traverse many times the same AST or subsets of the same AST (e.g., the entire file and then maybe a specific method or a class within that file). For this reason, having something to make these traversal quicker could be useful.

The measurements by @gabriele-tomassetti seems to indicate that after the freezing the traversal is over 10 times faster (but freezing has a cost).

Could @gabriele-tomassetti measure also the time for a second traversal without freezing, to verify if reflection caches help us there?

Yes, I have created the logic to patch the bytecode to remove reflection, as I was trying to use the GraalVM (which was not too happy about reflection). We could potentially also explore that route. The pros is that it does not rely on the freezing mechanism, but the cons is that it requires an extra build step, which is not that trivial. It could be used only for our internal parsers, I think.

gabriele-tomassetti:
> Could @gabriele-tomassetti measure also the time for a second traversal without freezing, to verify if reflection caches help us there?

The time for a second traversal is basically unchanged:

```
Walk time: 2724
Walk time (second time): 2679
```

We should remember that the user could also cache the result of the walk themselves.

```
// ast is a node
val nodes = mutableListOf<Node>()
ast.walk().forEach { nodes.add(it) }
```

So, if we cannot provide a particularly smart and effective method to cache the AST, I am not sure how much makes sense to provide a cached walk. The only advantage would be for the user that does not know that the walk itself is costly in terms of performance.

alessiostalla:
> The `walk` method already uses a sequence.

It _produces_ a sequence, but it does so by creating intermediate lists that it throws away (each call to `children`). Changing it to avoid building these temporary lists gives:
```
Nodes: 699460
Walk time: 4041
Walk time frozen: 3697
```
I've pushed the "pure sequence" version to this branch if you want to further discuss or improve it.

gabriele-tomassetti:
> It _produces_ a sequence, but it does so by creating intermediate lists that it throws away (each call to `children`). Changing it to avoid building these temporary lists gives:

Sorry, you were right.
If I am not mistaken, yielding the child and using `walkChildren` rather than` walk` is equivalent and performs better.

```
Walk time: 437
Walk time (second): 456
Walk time frozen: 443
```

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
```diff
@@ -15,7 +15,8 @@ fun Node.walk(): Sequence<Node> {
     return sequence {
         yield(this@walk)
         this@walk.walkChildren().forEach {
-            yieldAll(it.walk())
+            yield(it)
+            yieldAll(it.walkChildren())
```
alessiostalla, line 19: Doesn't this stop at the second level of children?

gabriele-tomassetti: You are right. I forgot to adapt the test. So, I am  mistaken.

gabriele-tomassetti:
Is there any way to detect if the `children` of a node are not going to change? I think that if all the properties were read-only, all the collections were immutable and all the types were data classes, or had value properties, the children would never change. However, I am not sure if we can check for that.

ftomassetti:
> Is there any way to detect if the `children` of a node are not going to change?

I would say no, as in general properties can be mutable and we do not have a mechanism to be notified about changes. This is something we may want to have in Kolasu 2

gabriele-tomassetti:
I have updated the pull request and moved the cache to a `FastWalker`. This way should be clearer to the user when the caching system is used.

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
```diff
@@ -104,6 +105,17 @@ fun Node.walkChildren(): Sequence<Node> {
     }
 }
 
+fun Node.walkChildrenProperties(): Sequence<PropertyDescription> {
```
alessiostalla, line 108: This appears to be unused

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
```diff
@@ -153,3 +165,33 @@ fun <T> Node.searchByType(
...
+
+    private fun getChildren(child: Node): List<Node> {
+        return if (childrenMap.containsKey(child)) {
+            childrenMap[child]!!
+        } else {
+            childrenMap.put(child, child.walkChildren().toList())
+            childrenMap[child]!!
+        }
+    }
+    fun walk(root: Node = node): Sequence<Node> {
```
alessiostalla, line 185: This works, but it duplicates the walking algorithm, and constrains the user to this single walk algorithm (so e.g. no `walkLeavesFirst`, `walkAncestors` and so on).

I was thinking about adding an extension method to `Sequence<Node>` like:

```
fun <T: Node> Sequence<T>.cached(): Sequence<T> {
    return this.toCollection(mutableListOf()).asSequence()
}
```

Then, one could do e.g. 

```
val nodes = root.walkLeavesFirst().cached()
nodes.forEach { ... }
// ...
nodes.map { ... }
```

so you pay the cost of the traversal only once. With my implementation, you pay it eagerly when calling `cached()`, probably there's a nice way to make it lazy so you only pay it the first time you actually traverse the nodes.

core/src/test/kotlin/com/strumenta/kolasu/traversing/TraversingStructurallyTest.kt
```diff
@@ -83,6 +81,72 @@ internal class TraversingStructurallyTest {
...
+        }
+        val walkTimeFastThree = measureTimeMillis {
+            var countedNodes = 0
+            fw.walk().forEach { countedNodes++ }
+            countedNodesList.add(countedNodes)
+        }
+        countedNodesList.forEach {
+            assertEquals(nodes, it)
+        }
+        assert(walkTimeFast > walkTime)
```
alessiostalla, line 146: This assertion looks a bit fragile since the two times are similar usually and GC etc. may make `walkTime` slightly longer than `walkTimeFast`.

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
gabriele-tomassetti: If I understand it correctly, you are saying that the FastWalker should look more like this:

```
class FastWalker(val node: Node) {
    private val walkMap: WeakHashMap<Node, Sequence<Node>> = WeakHashMap<Node, Sequence<Node>>()


    fun walk(root: Node = node): Sequence<Node> {      
        return if (walkMap.containsKey(node)) {
            walkMap[node]!!
        } else {
            walkMap.put(node, walk(node).cached())
            walkMap[node]!!
        }
    }
}
```

Is that correct?

gabriele-tomassetti: If you are saying that FastWalker itself is a bit superflous, I agree that it is kinda true. Its purpose is essentially to cover a hidden drawback of kolasu. The FastWalker is simply a convenience for people that are not aware that the walk takes more time than expected because it uses reflection. So it essentially automates saving the result of the walk for the user. If you were aware of the issue, you could already achieve the same results manually, right now, simply by saving the result of the walk.

So, what I mean is that I do not think that this:

`val nodes = root.walk().cached()`

would do something different than this, right now:

`val nodes = root.walk()`

At least `FastWalker` does also speed up subsequent walks starting at other points of the tree.

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -141,6 +141,10 @@ open class Node() : Origin, Destination {
     final override fun toString(): String {
         return "${this.nodeType}(${properties.joinToString(", ") { "${it.name}=${it.valueToString()}" }})"
     }
+
+    internal companion object Freezer {
```
alessiostalla, line 145: This doesn't appear to be used anymore

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
alessiostalla: Well, using the FastWalker is still an explicit choice, so you have to be aware of the problem anyway.

Your other point makes a real difference. If the user's pattern is to walk the same tree repeatedly starting from different nodes, then something like FastWalker is necessary to speed up the traversal. To avoid having to code a different FastWalker for each walking strategy, we could re-engineer FastWalker to work on sequences rather than nodes, something like the following:

```
class FastWalker(val sequence: Sequence<Node>) {
    private val walks = mutableMapOf<Node, Sequence<Node>>()

    fun walk(root: Node): Sequence<Node> {
        return walks.computeIfAbsent(root) {
            val iterator = sequence.iterator()
            for (node in iterator) {
                if (node == root) {
                    val cached = iterator.asSequence().toList()
                    return@computeIfAbsent sequence {
                        yield(node)
                        yieldAll(cached)
                    }
                }
            }
            throw Error("Node not in sequence")
        }
    }
}
```

I've checked that the above passes your tests. You use it like this:

```
val fw = FastWalker(root.walk())
fw.walk(root).forEach { countedNodes++ } // Slow
fw.walk(root).forEach { countedNodes++ } // Fast
```

alessiostalla: Ah, but this would still be slow the first time your walk starting from another root, it only improves if you walk twice from the same root.

alessiostalla: So, I think we should keep FastWalker as it is, and document that if you want a different traversal algorithm than `walk()`, you'll have to code your own version of FastWalker.

ftomassetti:
@gabriele-tomassetti is this PR ready to merge? If so, can you merge it?

Issue #50, opened 2021-08-25
Allow maps in nodes


alessiostalla:
http://wiki.eclipse.org/EMF/FAQ#How_do_I_create_a_Map_in_EMF.3F

alessiostalla:
Our VB6 parser uses maps, so we need this if we want to export examples for [Strumenta Playground](https://playground.strumenta.com).

ftomassetti:
I am moving this to version 2.0, as we may end up dropping support for EMF

ftomassetti:
We are dropping EMF, closing

PR #166, opened 2023-02-07
Ecore improvements

alessiostalla:
- Support GenericNode instances in Ecore models, otherwise, trying to serialize an AST that happens to contain GenericNodes throws an exception.
- Allow maps in nodes (#50 ). Note: this is limited to Ecore, with the sole purpose of exporting VB6 examples to Playground. No testing of `walk` etc. since the VB6 parser isn't using that.

### DIFF
--- a/emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
+++ b/emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
@@ -45,7 +45,7 @@ fun EClass.addReference(name: String, type: EClass, min: Int, max: Int): ERefere
     return eReference
 }
 
-fun EClass.addAttribute(name: String, type: EDataType, min: Int, max: Int): EAttribute {
+fun EClass.addAttribute(name: String, type: EClassifier, min: Int, max: Int): EAttribute {
     val eAttribute = EcoreFactory.eINSTANCE.createEAttribute()
     eAttribute.name = name
     eAttribute.eType = type
--- a/emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
+++ b/emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
@@ -45,7 +45,7 @@ fun EClass.addReference(name: String, type: EClass, min: Int, max: Int): ERefere
     return eReference
 }
 
-fun EClass.addAttribute(name: String, type: EClassifier, min: Int, max: Int): EAttribute {
+fun EClass.addAttribute(name: String, type: EDataType, min: Int, max: Int): EAttribute {
     val eAttribute = EcoreFactory.eINSTANCE.createEAttribute()
     eAttribute.name = name
     eAttribute.eType = type
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Processing.kt
@@ -311,8 +311,7 @@ fun Node.transformChildren(operation: (Node) -> Node) {
 fun Node.mapChildren(operation: (Node) -> Node): Node {
     val changes = mutableMapOf<String, Any>()
     relevantMemberProperties().forEach { property ->
-        val value = property.get(this)
-        when (value) {
+        when (val value = property.get(this)) {
             is Node -> {
                 val newValue = operation(value)
                 if (newValue != value) {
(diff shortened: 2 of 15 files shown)


emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
```diff
@@ -45,7 +45,7 @@ fun EClass.addReference(name: String, type: EClass, min: Int, max: Int): ERefere
     return eReference
 }
 
-fun EClass.addAttribute(name: String, type: EDataType, min: Int, max: Int): EAttribute {
+fun EClass.addAttribute(name: String, type: EClassifier, min: Int, max: Int): EAttribute {
```
ftomassetti, line 48: Why do we want attribute that are not simple data types? Is this to support maps?

alessiostalla: Yes, here http://wiki.eclipse.org/EMF/FAQ#How_do_I_create_a_Map_in_EMF.3F they recommend to define a Map$Entry subclass with a key and a value attributes, value would be a Node in our case.

alessiostalla:
> I reflected a little bit on this, and this seems an interesting problem.
> 
> Some issues I see:
> 
>     * We should restrict what can be used as key, probably only to something which is not a Node, otherwise we should consider children also in nodes

Yes, this only supports keys that are not nodes, but it doesn't enforce that in any way. We could throw an exception if we detect that situation.

>     * So far we had a predictable traversal order, but if we introduce maps, what is the traversal order?

I have no idea :) I left `walk` etc. out of scope. The VB6 parser already has maps in its AST; I believe that, even though they contain nodes, they aren't considered in walks etc. with the released version of Kolasu. Here I've set `providesNodes` to true for maps that have nodes as values, but maybe it's better to play it safe and, for now, consider maps as attributes, not children.
 
>     * This is a significant feature that would affect basically everything: from traversal, to serialization
> 
>     * This would prevent compatibility with LionWeb or MPS or other modeling systems as they do not support maps (just individual values and lists)

Strictly speaking, Ecore doesn't support maps either – it just suggests a well-known encoding that works within EMF. With this PR, when exporting to Ecore, maps are essentially translated into lists of pairs (the suggested encoding). In fact, this PR is only half of the work, the other half would be to detect that encoding and reconstruct a map in Tylasu for the playground.

> 
> 
> So my basic question is: do we see a really strong benefit for supporting maps?

Not crashing the export to Ecore when there is a map in a node was the only goal of this change, which is 100% of our VB6 ASTs. There's nothing in Kolasu preventing users to store maps in their nodes.

I think (but I would need to check that) that users can store composite objects that are not Nodes in their AST and we support serializing them, so excluding just maps looks a bit arbitrary, especially since maps are used relatively often.

> 
> **ADDED**: this is if we want maps to store nodes, if we want to put any other value than I think all my concerns go away

Unfortunately, the VB6 parser has properties that are maps of strings to Node instances.

alessiostalla:
Also, this PR should have been split in two: one for GenericNodes and the other for maps. Unfortunately I mistakenly pushed both changes to the same branch.

alessiostalla:
Addendum: even without maps, traversal order is already unpredictable if the AST contains sets, I've checked it.

ftomassetti:
I think this may be a topic to cover in a technical meeting or during the sync. What do you think?

ftomassetti:
We discussed this at the last technical meeting, but I cannot recall if we reached a decision and which one was it.
Would it be useful to have a call between the two of us and get this solved one way or another?

alessiostalla:
IIRC we decided not to support maps, and that the VB6 parser will at some point be re-engineered to avoid including maps in the AST.

ftomassetti:
Thank you Alessio. I then created an issue in VB6. Sorry for the effort you put into this, but I think this lead to a very useful conversation.

Could it be possible to revise the PR to keep the part about the GenericNode?

alessiostalla:
Yes, absolutely. It was a mistake to mix the two in the first place.

emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
ftomassetti: I think here we can revert back to EDataType

ftomassetti (approved):
+1

emf/src/main/kotlin/com/strumenta/kolasu/emf/metamodel_building.kt
ftomassetti: @alessiostalla can we revert this part back to EDataType and merge this PR?

PR #171, opened 2023-03-08
Feature: Kolasu lexer interface

alessiostalla:


core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
```diff
@@ -60,26 +60,10 @@ abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext> : ASTPa
...
-    /**
-     * Performs "lexing" on the given code stream, i.e., it breaks it into tokens.
-     */
-    @JvmOverloads
-    fun lex(
-        inputStream: InputStream,
-        charset: Charset = Charsets.UTF_8,
-        onlyFromDefaultChannel: Boolean = true
-    ): LexingResult {
+    override fun lex(inputStream: InputStream, charset: Charset, onlyFromDefaultChannel: Boolean): LexingResult {
```
ftomassetti, line 63: It seems to me that we lost the default values, the comment, and the @JvmOverloads annotation (useful for using this from Java)

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -36,12 +36,27 @@ open class CodeProcessingResult<D>(
...
+        val KEYWORD = TokenCategory("Keyword")
+        val NUMERIC_LITERAL = TokenCategory("Numeric literal")
+        val STRING_LITERAL = TokenCategory("String literal")
+        val PLAIN_TEXT = TokenCategory("Plain text")
+    }
+}
+
+open class KolasuToken(open val category: TokenCategory, open val position: Position)
+
+data class KolasuANTLRToken(override val category: TokenCategory, val token: Token) :
```
ftomassetti, line 51: Is this an adapter from the ANTLR Token to the KolasuToken? Maybe we could add a comment to specify that

core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
alessiostalla: I moved those to the interface.
Unfortunately, in Kotlin, interfaces can't have `@JvmOverloads`, and implementation cannot either, so I had to manually add all the overloads.

ftomassetti: But maybe this is the case as we are overriding?

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -119,12 +134,42 @@ class ParsingResult<RootNode : Node>(
...
 
-interface KLexer {
-    fun lex(code: String): LexingResult = lex(code.toStream())
-
+interface KolasuLexer {
+
+    /**
+     * Performs "lexing" on the given code string, i.e., it breaks it into tokens.
+     */
+    fun lex(code: String, onlyFromDefaultChannel: Boolean = true) =
```
ftomassetti, line 142: Can we use @JvmOverloads here?

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -119,12 +134,42 @@ class ParsingResult<RootNode : Node>(
...
+
+    /**
+     * Performs "lexing" on the given code string, i.e., it breaks it into tokens.
+     */
+    fun lex(code: String) = lex(code, true)
+
+    /**
+     * Performs "lexing" on the given code stream, i.e., it breaks it into tokens.
+     */
+    fun lex(
```
ftomassetti, line 153: Can we use @JvmOverloads here?

ftomassetti:
+1 with minor comments. Please feel free to consider or ignore the comments and merge

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: No, because it's an interface

alessiostalla:
I've added documentation, some testing, and refactored the lexing parts into a dedicated class.

PR #175, opened 2023-03-23
Expose ANTLR tokens in the default ANTLR-based lexer. 

### LINKED ISSUE #174: Adding TokenType to KolasuToken
In the RPG Parser we are facing this error:
Screenshot 2023-03-23 at 08 35 37

This is because now the lex method returns a KolasuToken and not an ANTLR's Token.
The KolasuToken does not have a type.

I think it could be useful to have KolasuTokens having a field type of type TokenType. A TokenType could be identified by a name and belong to a TokenCategory. While the lexer would only need the TokenCategory, other applications could need to distinguish between different token types within the same category, so I think that having this distinction could be valuable.

What @alessiostalla and @gabriele-tomassetti think about this?

alessiostalla:
Fixes #174.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
@@ -16,7 +16,7 @@ import java.util.*
 import kotlin.reflect.full.memberFunctions
 import kotlin.system.measureTimeMillis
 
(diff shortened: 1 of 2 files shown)


alessiostalla:
Not language-specific, but parser-specific, in case someone wants to integrate a parser that is not ANTLR maybe.

ftomassetti:
Ok, makes sense

PR #179, opened 2023-04-12
Make KolasuLexer parametric w.r.t. the type of token produced

ftomassetti:
These changes are required to make the RPG Lexer works, as it uses the RPGToken and not KolasuANTLRToken.
In my opinion this seems complicated, and we may consider adding the text field into KolasuToken, making it a concrete class. We could then make a method that given an ANTLR token convert it to a KolasuToken.


alessiostalla (approved):
Yeah it looks a bit too complicated but if it's necessary...

ftomassetti:
Thank you!

PR #180, opened 2023-04-13
Add support for RMI

ftomassetti:
By making elements Serializable we can expose the parsers through RMI, as we do in the Audit Viewer

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -1,12 +1,13 @@
 package com.strumenta.kolasu.model
 
+import java.io.Serializable
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.memberProperties
 
-interface Origin {
+interface Origin : Serializable {
     val position: Position?
     val sourceText: String?
     val source: Source?
@@ -21,7 +22,7 @@ data class CompositeOrigin(
     override val sourceText: String?
 ) : Origin
 
-interface Destination
+interface Destination : Serializable
 
 data class CompositeDestination(val elements: List<Destination>) : Destination
 data class TextFileDestination(val position: Position?) : Destination
@@ -32,7 +33,7 @@ data class TextFileDestination(val position: Position?) : Destination
  * It implements Origin as it could be the source of a AST-to-AST transformation, so the node itself can be
  * the Origin of another node.
  */
-open class Node() : Origin, Destination {
+open class Node() : Origin, Destination, Serializable {
 
     @Internal
     protected var positionOverride: Position? = null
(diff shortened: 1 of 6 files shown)


core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -1,12 +1,13 @@
...
 
+import java.io.Serializable
 import kotlin.reflect.KClass
 import kotlin.reflect.KProperty1
 import kotlin.reflect.KVisibility
 import kotlin.reflect.full.findAnnotation
 import kotlin.reflect.full.memberProperties
 
-interface Origin {
+interface Origin : Serializable {
```
alessiostalla, line 10: Is this necessary? Not all Origins will be serializable...

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -21,7 +22,7 @@ data class CompositeOrigin(
     override val sourceText: String?
 ) : Origin
 
-interface Destination
+interface Destination : Serializable
```
alessiostalla, line 25: Same here

alessiostalla (commented):
In general, I see no problem in making core data classes (e.g. Position) serializable.
With nodes, I'm a bit afraid of future compatibility problems. If we later implement AST nodes on top of LionWeb nodes, are we sure that everything in LionWeb is serializable, and that it doesn't drag all of the metamodels into the object graph?

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
ftomassetti: Yes, but a Node could specify an Origin or a Destination, so all Origins and all Destinations should be Serializable

alessiostalla: ParseTreeOrigin isn't, though, for example. IMHO it's more appropriate if only actual serializable concrete classes implement Serializable.

ftomassetti:
> In general, I see no problem in making core data classes (e.g. Position) serializable. With nodes, I'm a bit afraid of future compatibility problems. If we later implement AST nodes on top of LionWeb nodes, are we sure that everything in LionWeb is serializable, and that it doesn't drag all of the metamodels into the object graph?

This is a good question. I think that if our nodes extends the lionweb nodes and we send our nodes only and our nodes are serializable then it should work. For the metamodel classes... yes, they may also need to be serializable. Perhaps I could do that change on the LionWeb side, when needed. It is just about implementing a marker interface, so it should not cause any specific problem. We could do that if at that stage RMI is not presenting problems and we want to confirm we want to use it

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
ftomassetti: You are right, I make only specific subclasses Serializable now

ftomassetti:
@alessiostalla I addressed the comments. Would this PR now be acceptable?

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
ftomassetti: Changed

alessiostalla (approved):

PR #181, opened 2023-04-18
TrivialFactoryOfParseTreeToASTNodeFactory refactoring

ftomassetti:
Minor changes to the TrivialFactoryOfParseTreeToASTNodeFactory class, based on the work done in the Java language module.

Essentially:
* When we are waiting for a String or a String? we just get the text of the corresponding element, which could be not just a token but also a node such as IdentifierContext in the case of the Java language module
* When there is more than one constructor we pick the primary one, if available


alessiostalla (approved):

ftomassetti:
Thank you!

PR #183, opened 2023-04-18
Making the core module NOT depend on ANTLR

### LINKED ISSUE #105: Consider moving ANTLR out of the core module

ftomassetti:
This will permit in the future to implement parsers or lexers using something different from ANTLR. I basically removed the dependency from ANTLR in core and moved all the classes that did not compile to the new `antlr` module. It required very few changes. I also renamed the packages involving antlr to include the word `antlr`

Fix #105 

Please note that this is intended to be merged in the Kolasu 1.6 branch, not in the master branch.


loradd:
What about ANTLR parse tree nodes as `origin`? I might be missing something here, are we using `Origin` nodes and setting the text property or some kind of ANTLR node wrapper?

ftomassetti:
> What about ANTLR parse tree nodes as origin? I might be missing something here, are we using Origin nodes and setting the text property or some kind of ANTLR node wrapper?

`ParseTreeOrigin` moved to the ANTLR module. When using an ANTLR parser everything should work as before this PR: one would just need to include also the kolasu-antlr module. However if someone build a Kolasu node without using ANTLR then it will need to provide another subclass of Origin. Does this make sense?

loradd:
Yes, it makes sense to me! Nice :)

alessiostalla:
Makes sense & it's also roughly the process we've followed in Tylasu (though there it's still a single module with multiple entry points).

loradd (approved):
Looks good!

ftomassetti:
Thank you!

PR #185, opened 2023-04-18
Symbol resolution experiment

loradd:
Adds support for declaratively define symbol resolvers, i.e. `DeclarativeSymbolSolver`.


core/src/main/kotlin/com/strumenta/kolasu/symbolresolution/Symbols.kt
```diff
@@ -0,0 +1,6 @@
+package com.strumenta.kolasu.symbolresolution
+
+import com.strumenta.kolasu.model.PossiblyNamed
+
+interface Symbol : PossiblyNamed
```
ftomassetti, line 5: Note: I think that in Kolasu 2.0 a symbol could become a LIonWeb's Node (but not necessarily a Kolasu ASTNode, which will implement Node)

core/src/main/kotlin/com/strumenta/kolasu/utils/Memoization.kt
```diff
@@ -0,0 +1,10 @@
+package com.strumenta.kolasu.utils
+
+class Memoize1<in T, out R>(val f: (T) -> R) : (T) -> R {
```
ftomassetti, line 3: Why `Memoize1` is it because it takes one parameter?

core/src/test/kotlin/com/strumenta/kolasu/symbolresolution/DeclarativeSymbolResolverTest.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    scopeFor(OperationDecl::returns) { compilationUnit: CompilationUnit ->
+        Scope().apply { compilationUnit.content.forEach { this.define(it) } }
+    }
+}
+
+class SymbolResolutionTest {
+
+    @Test
+    fun example() {
```
ftomassetti, line 147: Good test!

ftomassetti (approved):
+1 Great to have this in Kolasu! 
P.S. Please feel free to go ahead and merge

core/src/main/kotlin/com/strumenta/kolasu/symbolresolution/Symbols.kt
loradd: Make sense! I think we should also think about extending the current symbol resolution mechanisms to support non-name-based symbol references, e.g. unnamed predicated references like 'the nearest enclosing class' or similar.

PR #186, opened 2023-04-18
Refactor token-related operations into a separate TokenFactory

alessiostalla:
Refactor token-related operations into a separate TokenFactory to be injected into parsers (with a default implementation for parsers wrapping ANTLR tokens).

This spares KolasuParser implementations from having to re-implement `lex` and other methods they couldn't inherit from KolasuANTLRLexer.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
@@ -45,7 +45,7 @@ interface TokenFactory<T : KolasuToken> {
     }
 }
 
-class DefaultTokenFactory : TokenFactory<KolasuANTLRToken> {
+class ANTLRTokenFactory : TokenFactory<KolasuANTLRToken> {
     override fun convertToken(t: Token): KolasuANTLRToken = KolasuANTLRToken(categoryOf(t), t)
 }
 
--- a/core/src/test/kotlin/com/strumenta/kolasu/parsing/KolasuParserTest.kt
+++ b/core/src/test/kotlin/com/strumenta/kolasu/parsing/KolasuParserTest.kt
@@ -12,7 +12,7 @@ import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
 
 class SimpleLangKolasuParser : KolasuParser<Node, SimpleLangParser, SimpleLangParser.CompilationUnitContext,
-    KolasuANTLRToken>(DefaultTokenFactory()) {
+    KolasuANTLRToken>(ANTLRTokenFactory()) {
     override fun createANTLRLexer(charStream: CharStream): Lexer {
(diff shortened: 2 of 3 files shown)


core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
```diff
@@ -16,7 +16,40 @@ import java.util.*
...
+        return if (ptRoot != null) {
+            extractTokensFromParseTree(ptRoot)
+            antlrTerminals.sortBy { it.symbol.tokenIndex }
+            val tokens = antlrTerminals.map { convertToken(it) }.toMutableList()
+            LexingResult(result.issues, tokens, result.code, result.firstStage.lexingTime)
+        } else null
+    }
+}
+
+class DefaultTokenFactory : TokenFactory<KolasuANTLRToken> {
```
ftomassetti, line 48: Could we call it ANTLRTokenFactory or DefaultANTLRTokenFactory?

core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
```diff
@@ -74,8 +105,8 @@ abstract class KolasuANTLRLexer : KolasuLexer<KolasuANTLRToken> {
  *
  * You should extend this class to implement the parts that are specific to your language.
  */
-abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext, T : KolasuToken> :
-    KolasuLexer<T>,
+abstract class KolasuParser<R : Node, P : Parser, C : ParserRuleContext, T : KolasuToken>(tokenFactory: TokenFactory<T>) :
+    KolasuANTLRLexer<T>(tokenFactory),
```
ftomassetti, line 109: Note that this was renamed in KolasuANTLRParser in 1.6, so it makes sense that here implements KolasuANTLRLexer and not KolasuLexer

ftomassetti: Should this get a default value (an instance of the DefaultTokenFactory)?

ftomassetti:
This seems a very good idea

ftomassetti (approved):
+1

core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
alessiostalla: It would be great if it had a default value, but it wouldn't fulfill its generics signature

core/src/main/kotlin/com/strumenta/kolasu/parsing/KolasuParser.kt
alessiostalla: Sure!

PR #188, opened 2023-04-18
Fix serialization of Serializable

### LINKED ISSUE #187: EMF test failing because Pair is no longer Serializable
Probably the EMF export removes all instances of `Serializable` but one of our tests expected it to be there.

ftomassetti:
Fix #187

### DIFF
--- a/emf/src/test/kotlin/com/strumenta/kolasu/emf/KolasuMetamodelTest.kt
+++ b/emf/src/test/kotlin/com/strumenta/kolasu/emf/KolasuMetamodelTest.kt
@@ -36,12 +36,6 @@ class KolasuMetamodelTest {
 
         val resource = createResource(mmuri) ?: throw IOException("Unsupported destination: $mmuri")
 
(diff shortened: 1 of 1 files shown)


emf/src/test/kotlin/com/strumenta/kolasu/emf/KolasuMetamodelTest.kt
```diff
@@ -40,7 +40,6 @@ class KolasuMetamodelTest {
             "java.io", "https://strumenta.com/javaio", "javaio",
             resource
         )
-        javaIOMetamodelBuilder.provideClass(java.io.Serializable::class)
         javaIOMetamodelBuilder.generate()
```
alessiostalla, line 43: Does it still make sense to include a package for java.io that doesn't contain any classes in the StarLasu metamodel?

ftomassetti: You are right, we should remove it

PR #189, opened 2023-04-18
Rename Position into Range

ftomassetti:
This touched a lot of files :)
I also updated the version number of the StarLasu metamodel


ftomassetti:
Note: maintain EMF metamodel unmutated, so that we do not break Playground

ftomassetti:
This one is a huge (but boring) change. It has been open for a couple of weeks and I will now proceed to merge it

PR #193, opened 2023-04-20
Adding RuleContext.ancestor method

ftomassetti:
This is convenient to navigate the parse tree during mapping


loradd (approved):
Looks good to me!

PR #204, opened 2023-05-29
Issue62

### LINKED ISSUE #62: Ensure properties are not duplicated
There are cases in which we override properties and we end up with two properties definition.

In parser bench this is visible:
Screenshot 2022-02-23 at 16 12 19

ftomassetti:
Fix #62 . As @alessiostalla suspected it was an EMF issue


alessiostalla (approved):
Makes sense, good find.

ftomassetti:
Thank you!

PR #211, opened 2023-05-31
Introduce transformations from one element to multiple nodes

### LINKED ISSUE #210: Verify if transformers can return lists of nodes and not only single nodes
See https://github.com/Strumenta/rpg-to-python-transpiler/issues/6

ftomassetti:
We may have this particular need when mapping one statement to many. 

I think it would be useful to have a review both from @alessiostalla , as the designer of the transformations mechanism, and from @mctaverna as the person who raised the need for this in https://github.com/Strumenta/rpg-to-python-transpiler/issues/6

There is only one example of usage in ASTTransformerTest.testTransforingOneNodeToMany, but I can add more based on your requests

Fix #210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,4 @@ generated-test-src
 /*.json
 /*.xmi
 **/bin/
+.netlify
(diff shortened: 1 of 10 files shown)


ftomassetti:
@alessiostalla you are right, I revised the name of the methods so that `transform` has the same signature as before. The other method is now called `transformIntoNodes`.

@mctaverna I can also publish the snapshot, so you can test it in the RPG to Python transpiler project and check if it works for your needs

ftomassetti:
@mctaverna what do you think about this?

mctaverna:
@ftomassetti works for me, should be integrated into the RPGParser?

ftomassetti:
@mctaverna  thank you! Yes, please. I will release a new version of Kolasu with this change

PR #213, opened 2023-05-31
Reorganizing range, source, origin, and destinations

### LINKED ISSUE #194: Revisit relation between Origin, Range (previously Position), Source
For example, in some nodes I set node.position.source and then node.source returns something else

ftomassetti:
The node itself now contains exclusively the data but not the position or the source. They are part of the origin of the node.
The node has accessors to simplify getting access to them.

- rangeOverride is moved to the ParseTreeOrigin.

- detach is renamed in detachFromParseTree, because detach could be interpreted as "remove the node from the AST" (i.e., set its parent to null.

- destinations and origins are moved to separate files

- the Node is now not anymore an Origin or a Destination, both still it can be used almost as an Origin or a Destination by means of a few operators being added in Destinations.kt and overloaded methods such as `Node.withOrigin(Node)`

Fix #194

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 version=1.6.0-RC1-observers-SNAPSHOT
 kotlin_version=1.8.21
 dokka_version=1.8.10
-antlr_version=4.12.0
+antlr_version=4.9.3
 clikt_version=3.5.2
 org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
 jvm_version=1.8
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,7 +1,7 @@
 version=1.6.0-RC1-observers-SNAPSHOT
 kotlin_version=1.8.21
 dokka_version=1.8.10
-antlr_version=4.9.3
+antlr_version=4.12.0
 clikt_version=3.5.2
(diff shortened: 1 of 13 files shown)


alessiostalla (commented):
What's the impact on the EMF metamodel?

gradle.properties
```diff
@@ -1,7 +1,7 @@
 version=1.6.0-RC1-observers-SNAPSHOT
 kotlin_version=1.8.21
 dokka_version=1.8.10
-antlr_version=4.9.3
+antlr_version=4.12.0
```
alessiostalla, line 4: This will make Java 11+ a requirement, not that it's wrong but we have to be aware of that

ftomassetti: Right, let's revert that

ftomassetti:
> What's the impact on the EMF metamodel?

Good question. I think that we were already generating EObjects to wrap the Nodes and adapt them to be Origins and Destinations (i.e., we already had the EClasses `NodeOrigin` and `NodeDestination`), so the impact should be none, I think. By that I mean that the EMF generated would be the same

alessiostalla (approved):
It looks good to me. The impact on test ASTs seems minimal, so I don't expect this causing particular problems when migrating

ftomassetti:
Thank you @alessiostalla !

PR #214, opened 2023-06-01
Annotations

ftomassetti:
With this PR we add annotations.

Annotations are useful to attach data to Nodes successively to their creation, without having to change the AST class and "polluting it" with data useful only for certain elaborations.

Given this is a new significant feature, an opinion from @alessiostalla would be useful

Related to #203


core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -130,6 +134,47 @@ open class Node() : Serializable {
...
+            null
+        } else if (instances.size == 1) {
+            instances.first()
+        } else {
+            throw IllegalStateException("More than one instance of $kClass found")
+        }
+    }
+
+    fun <A : Annotation>addAnnotation(annotation: A): A {
+        if (annotation.annotatedNode != null) {
```
alessiostalla, line 158: What's the rationale for tracking the node an annotation is attached to?

alessiostalla (approved):
Looks good to me. As always, I'm raising a flag regarding exports – do we want to include annotations in EMF models, JSON exports, etc.? If we want, but it's out of scope, we should keep track with an issue.

ftomassetti:
> Looks good to me. As always, I'm raising a flag regarding exports – do we want to include annotations in EMF models, JSON exports, etc.? If we want, but it's out of scope, we should keep track with an issue.

You are right. I will add issues for this, and we can proceed with the implementation once we have used annotations and we are sure they are working well. Otherwise the risk is modifying them and then havin

Issue #200, opened 2023-05-29
Revisit observers


ftomassetti:
- [ ] consider using RxKotlin
- [ ] consider using a library for mutable lists
- [ ] consider using WeakReference in observers
- [ ] measure performance on a large AST

ftomassetti:
@alessiostalla https://github.com/ReactiveX/RxKotlin seems to be abandoned (last commit was 3 years ago). RxJava is still maintained, but I am reading online that people are looking at the flow APIs of Kotlin (something I guess is coming from the Kotlin standard library). What do you think?

alessiostalla:
I'm biased in favor of Rx* because you can use the same concepts and similar APIs in multiple languages, so we could more easily port observers to Tylasu, Pylasu etc.
But I don't know the flow APIs, so I can't really compare them.

PR #219, opened 2023-06-05
Using RxJava to implement observers

ftomassetti:
We do not use RxKotlin, as it seems abandoned, but we use RxJava, as suggested by @alessiostalla


ftomassetti:
This is part of #200

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
```diff
@@ -22,21 +25,31 @@ interface Named : PossiblyNamed {
...
  * This is not statically enforced as we may want to use some interface, which cannot extend Node.
  * However, this is enforced dynamically.
  */
-class ReferenceByName<N>(val name: String, initialReferred: N? = null) : Serializable where N : PossiblyNamed {
+class ReferenceByName<N>(val name: String, initialReferred: N? = null) :
+    Serializable,
+    Disposable,
+    ObservableSource<ReferenceChangeNotification<N>> where N : PossiblyNamed {
+
+    private val observers: MutableList<Observer<in ReferenceChangeNotification<N>>> = mutableListOf()
```
alessiostalla, line 42: Usually in situations like this I've employed a [Subject](https://github.com/ReactiveX/RxJava/wiki/Subject) instead of managing a list of observers manually.
Either a `BehaviorSubject` (if you want new observers to be notified of the last changed value upon subscription) or a `PublishSubject` (if you want new observers to only see changes after they've subscribed).

In Angular, the convention is to append a dollar sign to the subject, e.g.
```
val changes$ = PublishSubject<...>()
```
but I don't know if that's an Angular-specific convention or if it's common across Rx* libraries.

Either way, subscribers would access the public property e.g.

```
myReference.changes.subscribe(...)
```

Then, you don't have to manage the observers and call `onNext` on each, you can just call `onNext` on the Subject. See also https://github.com/ManyDesigns/Portofino/blob/deaba48d93502c1610b830781fa78855e6e05160/portofino-database/src/main/java/com/manydesigns/portofino/persistence/Persistence.java#LL106C45-L106C45 for an example using RxJava.

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ObservableList.kt
```diff
@@ -1,16 +1,19 @@
...
-        observers.add(observer)
-    }
+data class ListAddition<E>(val added: E) : ListNotification<E>()
+data class ListRemoval<E>(val removed: E) : ListNotification<E>()
+
+class ObservableList<E>(private val base: MutableList<E> = mutableListOf()) :
+    MutableList<E> by base,
+    ObservableSource<ListNotification<E>>,
+    Disposable {
+    private val observers = mutableListOf<Observer<in ListNotification<E>>>()
```
alessiostalla, line 16: Same here, I would use a public Subject property

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ReferenceToNodeObserver.kt
```diff
@@ -0,0 +1,44 @@
+package com.strumenta.kolasu.model.observable
+
+import com.strumenta.kolasu.model.Node
+import com.strumenta.kolasu.model.PossiblyNamed
+import com.strumenta.kolasu.model.ReferenceChangeNotification
+import io.reactivex.rxjava3.core.Observer
+import io.reactivex.rxjava3.disposables.Disposable
+
+class ReferenceToNodeObserver<N : PossiblyNamed>(val container: Node, val referenceName: String) :
```
alessiostalla, line 9: I don't understand the purpose of this class very well, but I have the sensation this should not be an observer. An observer acts as a door out of the Rx framework, and calling `onNext` from an observer directly usually is a code smell. The idiomatic way to do this would be to use an operator such as [map](https://github.com/ReactiveX/RxJava/wiki/Transforming-Observables#map) if the purpose is to transform events from one type to the other. E.g. something like

```
ref.changes.map(...).subscribe(node.notifications)
```

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
```diff
@@ -1,16 +1,25 @@
...
+typealias NodeObserver = Observer<in NodeNotification<in Node>>
+
 /**
  * The Abstract Syntax Tree will be constituted by instances of Node.
  */
-open class Node() : Serializable {
+open class Node() : Serializable, ObservableSource<NodeNotification<in Node>>, Disposable {
+
+    @property:Internal
+    val observers: MutableList<NodeObserver> = mutableListOf()
```
alessiostalla, line 22: Also here I would use something like
```
val notifications = PublishSubject<NodeNotification<in Node>>.create()
```

alessiostalla (changes requested):
I think there are a few code smells / non-idiomatic patterns. Also not calling dispose (i.e. unsubscribing) can be a source of memory leaks, so it should be clearly defined (and preferably tested) how one is supposed to unsubscribe.

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
ftomassetti: Implemented, please let me know if I did correctly

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ObservableList.kt
ftomassetti: Implemented, please let me know if I did correctly

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
ftomassetti: Implemented, please let me know if I did correctly.

I called it `changes` as in the other cases. Should I call it `notifications` here instead?

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ReferenceToNodeObserver.kt
ftomassetti: The idea is to use it to propagate the changes on the references to the node, in a convenient way. For example:
```
data class NodeWithReference(val ref: ReferenceByName<NamedNode>, val id: Int) : Node() {
    init {
        ref.subscribe(ReferenceToNodeObserver(this, "ref"))
    }
}
```
This code should either written manually or injected by the compiler plugin through a cumbersome API, so the simpler is the call, the better

ftomassetti: I understand I should change the example in ObservableNodeTest to something like:

```
data class NodeWithReference(val ref: ReferenceByName<NamedNode>, val id: Int) : Node() {
    init {
        //ref.subscribe(ReferenceToNodeObserver(this, "ref"))
        ref.changes.map(...).subscribe(this.changes)
    }
}
```

But it is not clear to me what should I specify where the dots are. Could you help me on this?

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
```diff
@@ -229,4 +225,8 @@ open class Node() : Serializable {
         val ref: ReferenceByName<T> = getReference(referenceName)
         ref.referred = referred
     }
+
+    fun subscribe(observer: Observer<NodeNotification<in Node>>) {
+        this.changes.subscribe(observer)
```
alessiostalla, line 230: Same here

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ReferenceToNodeObserver.kt
alessiostalla: Something like 

```
ref.changes.map { referenceNotification -> 
    return if (referenceNotification.oldValue != null) {
        ReferencedToRemoved(referenceNotification.oldValue, referenceName, container)
    } else if (referenceNotification.newValue != null) {
        ReferencedToAdded(referenceNotification.newValue, referenceName, container)
    } else ??? // I have no idea if this is possible/what it would represent, if the idea is that this should be dropped then the mergeMap operator is more adequate
}
```

alessiostalla (changes requested):
This looks way better imo, still a couple suggestions but I think we're close to something we can test in a broader set of cases

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ReferenceToNodeObserver.kt
ftomassetti: What we want to do is this:
* Always notify the owner of the reference about the change
* Notify the previous target of the reference (if it was not null) about the change
* Notify the next target of the reference (if it is not null) about the change       

For the first point I think we can write:
```
        ref.changes.map { referenceNotification ->
            ReferenceSet(
                container,
                referenceName,
                referenceNotification.oldValue as Node?,
                referenceNotification.newValue as Node?
            ) as NodeNotification<in Node>
        }.subscribe(this.changes)
```

However I do not know how to specify the other two cases, as in that case the thing that should receive the notification depend dynamically from the past and future value of the reference target.

You can see that in `ReferenceToNodeObserver`:

```
    override fun onNext(referenceNotification: ReferenceChangeNotification<N>) {
        container.changes.onNext(
            ReferenceSet(
                container,
                referenceName,
                referenceNotification.oldValue as Node?,
                referenceNotification.newValue as Node?
            )
        )
        if (referenceNotification.oldValue != null) {
            referenceNotification.oldValue.changes.onNext(
                ReferencedToRemoved(referenceNotification.oldValue, referenceName, container)
            )
        }
        if (referenceNotification.newValue != null) {
            referenceNotification.newValue.changes.onNext(
                ReferencedToAdded(referenceNotification.newValue, referenceName, container)
            )
        }
    }
```

Is this possible to do using map or with another call?            

I tried with this:

```
        ref.changes.mapOptional { referenceNotification ->
            if (referenceNotification.oldValue != null) {
                Optional.of(ReferencedToRemoved(referenceNotification.oldValue!!, referenceName, container))
            } else {
                Optional.empty()
            } as Optional<ReferencedToRemoved<in Node>>
        }.subscribe { referenceNotification ->
            (referenceNotification.node as Node).changes
        }
        ref.changes.mapOptional { referenceNotification ->
            if (referenceNotification.newValue != null) {
                Optional.of(ReferencedToAdded(referenceNotification.newValue!!, referenceName, container))
            } else {
                Optional.empty()
            } as Optional<ReferencedToAdded<in Node>>
        }.subscribe { referenceNotification ->
            (referenceNotification.node as Node).changes
        }
```

But it does not seem to work

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
```diff
@@ -68,11 +76,15 @@ class ReferenceByName<N>(val name: String, initialReferred: N? = null) : Seriali
 
         return true
     }
+
+    fun subscribe(observer: Observer<in ReferenceChangeNotification<N>>) {
+        changes.subscribe(observer)
```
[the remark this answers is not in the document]
ftomassetti: I checked but subscribe does not return anything in this case

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
ftomassetti: Screenshot 2023-06-19 at 17 14 03

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
alessiostalla: Changes is fine I think

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
alessiostalla: Oh OK I was only aware of `subscribe` overloads returning `Disposable`

core/src/main/kotlin/com/strumenta/kolasu/model/observable/ReferenceToNodeObserver.kt
ftomassetti: Thank you, your solution looks much better. I fixed the issue with the IR plugin. Basically it needs to generate for each AST class and each references it has this code:

```
init {
       ref.setContainer(this, "ref name")
}
```

alessiostalla (approved):
This looks good!

ftomassetti:
Thank you @alessiostalla for your help on this one. The result is much better than what we started with

PR #226, opened 2023-06-12
Symbol resolution API improvements

loradd:
This PR proposes the following improvements to the current Symbol Resolution API:

- Remove the `Symbol` interface and replace its usages with `PossiblyNamed` - this allows symbol resolution implementation to be more flexible, i.e. there is no need to tag nodes as `Symbols` anymore. This also fixes a problem with Ecore exporting as `Symbol` was not handled and hence gave issues during export (it would have required implementing an handler);
- Improves testing support - it makes testing reference resolution incremental, i.e. it is possible to limit tests on specific properties, classes (or all)

@ftomassetti 
@mctaverna 
@alessiostalla

What do you think?


core/src/main/kotlin/com/strumenta/kolasu/symbolresolution/Testing.kt
```diff
@@ -5,24 +5,57 @@ import com.strumenta.kolasu.model.PossiblyNamed
 import com.strumenta.kolasu.model.ReferenceByName
 import com.strumenta.kolasu.model.nodeProperties
 import com.strumenta.kolasu.traversing.walk
+import kotlin.reflect.KType
```
ftomassetti, line 8: This is very good as one of our goals is to facilitate testing of our parsers

core/src/main/kotlin/com/strumenta/kolasu/symbolresolution/Testing.kt
```diff
@@ -5,24 +5,57 @@ import com.strumenta.kolasu.model.PossiblyNamed
...
-                )
-            }
-            .forEach { property -> assertTrue { (property.get(it) as ReferenceByName<*>).resolved } }
+            .filter { property -> property.returnType.isSubtypeOf(referenceByNameType<T>()) }
+            .mapNotNull { property -> property.get(it) }
+            .map { value -> (value as ReferenceByName<*>).resolved }
     }
 }
+
+fun Node.getReferenceResolutionInfo(targetProperty: ReferenceByNameProperty): Sequence<Boolean> {
```
ftomassetti, line 48: I do not understand the goal of this operation by looking at its name

loradd: This is used to retrieve all values of the `resolved` field for each `ReferenceByName` of the given `targetProperty`. Actually, it is only used internally and not meant to be used outside. Perhaps, it would be better to make it private, what do you think? The other variations of this function might need to be inlined as currently can not be defined as private - the compiler would give error as private inline functions can not be used in public functions.

core/src/test/kotlin/com/strumenta/kolasu/symbolresolution/DeclarativeLocalSymbolResolverTest.kt
```diff
@@ -0,0 +1,224 @@
...
+) : StmtNode(), Named
+
+data class AssignmentStmt(
+    var lhs: ExprNode,
+    var rhs: ExprNode,
+) : StmtNode()
+
+sealed class ExprNode : Node(), Expression
+
+// a.v.c.d
```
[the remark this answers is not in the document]
loradd: Probably just a remainder to show what a `RefExpr` is supposed to mean, i.e. things like `a.b.c.d` where `a.b.c` would represent the context. I will remove it

loradd:
> Looks good to me, I just added minor comments

Thanks! I just pushed some changes to address the comments, thanks for the review!

PR #237, opened 2023-07-03
JSON generator: use nodeType

### LINKED ISSUE #235: JSON generator: use nodeType
Currently, it uses the Java class name but that doesn't work for wrapper/dynamic nodes.

alessiostalla:
Fixes #235

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
@@ -187,8 +187,9 @@ class JsonGenerator {
         withDestinationIds: IdentityHashMap<Node, String>? = null
     ):
         JsonElement {
+        val nodeType = node.nodeType
         val jsonObject = jsonObject(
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
```diff
@@ -187,8 +187,9 @@ class JsonGenerator {
         withDestinationIds: IdentityHashMap<Node, String>? = null
     ):
         JsonElement {
+        val nodeType = node.nodeType
         val jsonObject = jsonObject(
-            JSON_TYPE_KEY to if (shortClassNames) node.javaClass.simpleName else node.javaClass.canonicalName,
+            JSON_TYPE_KEY to if (shortClassNames) nodeType.substring(nodeType.indexOf('.') + 1) else nodeType,
```
ftomassetti, line 192: I think you want to find the last index of '.'. Alternatively I would use `nodeType.split('.').last()`

alessiostalla: Right, good catch!

PR #238, opened 2023-07-04
Supporting the recognition of internal classes when mapping to EMF

ftomassetti:
Without this change internal classes are considered to belong to a separate package. For example, a class Foo defined in a class a.b.Bar, will appear to belong to the package a.b.Bar. With this change it is recognized to belong to the package a.b, and it will have name Bar.Foo.

This is useful to generate documentation that looks like this:

Screenshot 2023-07-04 at 11 29 55


alessiostalla (approved):
Makes sense!

PR #242, opened 2023-07-27
Fix issue with transformation of children

### LINKED ISSUE #241: Problem in Transformation.kt?
Hi, I've tried to migrate kolasu-entity-parser to the last kolasu version  1.5.24.

I've migrated the code of `EntityParseTreeToAstTransformer (just changing the order of arguments)` but  the test keep failing with:
```
class com.strumenta.entity.parser.AntlrEntityParser$String_typeContext cannot be cast to class java.util.List (com.strumenta.entity.parser.AntlrEntityParser$String_typeContext is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')
java.lang.ClassCastException: class com.strumenta.entity.parser.AntlrEntityParser$String_typeContext cannot be cast to class java.util.List (com.strumenta.entity.parser.AntlrEntityParser$String_typeContext is in unnamed module of loader 'app'; java.util.List is in module java.base of loader 'bootstrap')
	at com.strumenta.kolasu.transformation.ASTTransformer.setChild(Transformation.kt:362)
```
But I can't figure out how to fix the mapping.

Thx

ftomassetti:
Fix #241


ftomassetti:
@alessiostalla would you have time to review this one this week?

alessiostalla:
Sure. I've pushed a commit refactoring and (hopefully) simplifying stuff a bit.
I didn't understand why a child node factory was expected to always return a list so I rewrote that part. It's a little embarrassing that no test caught that, so I made sure to include singular, optional, and multiple values in the test.

ftomassetti:
Thank you Alessio, adding the tests is a very good idea. I will check if that works with the entity example on which the issue was reported and if this is the case I will make a release

PR #247, opened 2023-08-11
Tentative fix to deal with issues using generateJSONWithStreaming

gabriele-tomassetti:
I have an issue with serializing JSON with streaming for the RPG Parser. The issue is that when generating JSON with streaming there is essentially an alternative serialization process that is not affected by custom serializers and works a bit differently. This tentative fix effectively uses for the serialization with streaming the same methods for normal serialization.

I have also ensured that it is always used the `gsonBuilder` with the custom serializers.

However, I do not know if there was a specific reason for using alternative serialization methods for streaming serialization. Maybe I am missing something.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonDeserializer.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonDeserializer.kt
@@ -41,7 +41,19 @@ class JsonDeserializer {
                     val actualClass = try {
                         Class.forName(className)
                     } catch (ex: ClassNotFoundException) {
-                        Class.forName(rawClass.`package`.name + "." + className)
+                        // This handles only the case when the serialized JSON used the simple name
+                        // rather than the canonical name for indicating the type of the serialized object.
+                        // The rawClass is the class we are expecting,
+                        // while className is the type in the JSON serialization.
+                        // These two types could be different: rawClass could be a sealed or abstract type,
+                        // while className could be a compatible concrete class.
+                        // So, when className is a simple name and we cannot resolve it directly,
+                        // we need to get the package name of the rawClass and apply it to className.
+                        Class.forName(
+                            "${rawClass.canonicalName.substring(
+                                0, rawClass.canonicalName.lastIndexOf('.') + 1
+                            )}$className"
+                        )
                     }
                     return deserialize(actualClass.asSubclass(Node::class.java), json.asJsonObject)
                 }
--- a/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonDeserializer.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonDeserializer.kt
@@ -38,7 +38,11 @@ class JsonDeserializer {
             when {
                 Node::class.java.isAssignableFrom(rawClass) -> {
                     val className = json.asJsonObject[JSON_TYPE_KEY].asString
-                    val actualClass = Class.forName(className)
+                    val actualClass = try {
+                        Class.forName(className)
+                    } catch (ex: ClassNotFoundException) {
+                        Class.forName(rawClass.`package`.name + "." + className)
+                    }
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonGenerator.kt
```diff
@@ -130,27 +131,28 @@ class JsonGenerator {
...
-            result.root.toJsonStreaming(writer, shortClassNames)
+            generateJSONWithStreaming(result.root, writer, shortClassNames)
         }
         writer.endObject()
     }
 
     fun generateJSONWithStreaming(root: Node, writer: JsonWriter, shortClassNames: Boolean = false) {
-        root.toJsonStreaming(writer, shortClassNames)
+        val gson = gsonBuilder.setPrettyPrinting().create()
+        gson.toJson(generateJSON(root, null, null, null, shortClassNames), writer)
```
ftomassetti, line 141: here we could specify the names of the parameters, to make this more understandable

core/src/test/kotlin/com/strumenta/kolasu/serialization/JsonGenerationTest.kt
```diff
@@ -115,10 +115,9 @@ class JsonGenerationTest {
         assertEquals(
             """{"#type":"com.strumenta.kolasu.serialization.MyRoot",
                 |"mainSection":{"#type":"com.strumenta.kolasu.serialization.Section","contents":
-                |[{"#type":"com.strumenta.kolasu.serialization.Content","annidatedContent":null,"id":1},
-                |{"#type":"com.strumenta.kolasu.serialization.Content","annidatedContent":
-                |{"#type":"com.strumenta.kolasu.serialization.Content","annidatedContent":
-                |{"#type":"com.strumenta.kolasu.serialization.Content","annidatedContent":null,"id":4},"id":3},"id":2}],
+                |[{"#type":"com.strumenta.kolasu.serialization.Content","id":1},
```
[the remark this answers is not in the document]
ftomassetti: It would also be useful to have tests to load the examples, both with shortClassNames set to false and to true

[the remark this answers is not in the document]
gabriele-tomassetti: Added more examples for generation, still working on deserialization tests.

core/src/main/kotlin/com/strumenta/kolasu/serialization/JsonDeserializer.kt
```diff
@@ -38,7 +38,11 @@ class JsonDeserializer {
             when {
                 Node::class.java.isAssignableFrom(rawClass) -> {
                     val className = json.asJsonObject[JSON_TYPE_KEY].asString
-                    val actualClass = Class.forName(className)
+                    val actualClass = try {
+                        Class.forName(className)
+                    } catch (ex: ClassNotFoundException) {
+                        Class.forName(rawClass.`package`.name + "." + className)
```
ftomassetti, line 44: Why?

ftomassetti: Also, does not exist a method that check if a class exist without throwing an exception?

gabriele-tomassetti: I was just about to ask your opinion on this fix. It was needed to pass this test [deserializeNodeFromStreamingWithShortNames](https://github.com/Strumenta/kolasu/blob/24807dacbf6516772cf004c9c94ed13014d98be2/core/src/test/kotlin/com/strumenta/kolasu/serialization/JsonDeserializerTest.kt#L87-L103). It seems that sometimes with simple names cannot find the right class.

I will check for that method.

gabriele-tomassetti: There seems to be no such method. There is `forName` overload that does not throw an exception, but is Java 9+.

image

PR #251, opened 2023-08-23
Add support for clearing caches

### LINKED ISSUE #250: Add mechanism to clean internal caches
In the Java module we do the following:
```
    override fun parse(
        file: File,
        charset: Charset,
        considerPosition: Boolean,
        measureLexingTime: Boolean
    ): ParsingResult<JCompilationUnit> {
        executionCounter++
        if (executionCounter % 500 == 0) {
            JavaLexer.cleanCache()
            JavaParser.cleanCache()
        }
        return super.parse(file, charset, considerPosition, measureLexingTime)
    }
```

Where in the JavaLexer we have:
```
	public static void cleanCache() {
			for (int d = 0; d < _decisionToDFA.length; d++) {
				_decisionToDFA[d] = new DFA(_ATN.getDecisionState(d), d);
			}
		}
```

And in the JavaParser we have:

```
	public static void cleanCache() {
			for (int d = 0; d < _decisionToDFA.length; d++) {
				_decisionToDFA[d] = new DFA(_ATN.getDecisionState(d), d);
			}
		}
```

We should find a way to enable this behavior by default

ftomassetti:
This mechanism will permit to avoid out of memory issues by automatically clearing the internal ANTLR cache after a certain number of executions.

Fix #250


ftomassetti:
While the changes for this fix are minimal, because of an issue with KtLint I added to upgraded and because of new styling rules there are now a ton of irrelevant changes

ftomassetti:
I suggest looking only at this commit: https://github.com/Strumenta/kolasu/pull/251/commits/2d07402d2497257a6d0886d93368ed207bd18ec5

ftomassetti:
In practice this PR will substitute the current solution which consists of adding this to both the lexer and the parser:
```
@members {
public static void cleanCache() {
		for (int d = 0; d < _decisionToDFA.length; d++) {
			_decisionToDFA[d] = new DFA(_ATN.getDecisionState(d), d);
		}
	}
}
```

And then adding logic to trigger the cleanCache methods from the subclass of KolasuParser. The current solution is also limited by the fact that some methods of KolasuParser are not open, so the solution is partial. This new solution should always work and require no internal knowledge of ANTLR and less work

alessiostalla:
I have two objections to this:

1. the method using reflection seems to do the same thing as ANTLR's built-in `clearCaches` methods on the `Parser` and `Lexer` class. Is there a reason for not using the built-in methods?
2. in my experience, clearing the DFA is not sufficient because ANTLR parsers also cache a shared (static) `PredictionContextCache` that they never clear. I've experimented this on the SAS parser where I've also put in place a solution that avoids sharing the `PredictionContextCache` statically thus allowing one to clear it.

ftomassetti:
Thank you for the feedback.

Regarding 1., I do not see the method. I have looked at: https://www.antlr.org/api/Java/org/antlr/v4/runtime/Parser.html
Regarding 2., great, do you think we can "steal" that strategy and generalize it so that it comes for free from using Kolasu?

alessiostalla:
Regarding 1. sorry they're actually on the [ATNSimulator class](https://www.antlr.org/api/Java/org/antlr/v4/runtime/atn/ATNSimulator.html) that you can access through the `Parser.interpreter` and `Lexer.interpreter` fields.
Regarding 2. yes sure we can, it's a general technique.

alessiostalla:
I'll add a commit adapting what I've got in the SAS parser.
Also I'm unconvinced about clearing the cache after a fixed numbers of iterations. In my experience, it's sufficient to encounter one "wrong" file to fill the caches excessively and slowing everything down. I think the policy for clearing the cache is too parser-specific to be hardcoded like that.

ftomassetti:
> I'll add a commit adapting what I've got in the SAS parser.

Thank you!

> Also I'm unconvinced about clearing the cache after a fixed numbers of iterations. In my experience, it's sufficient to encounter one "wrong" file to fill the caches excessively and slowing everything down. I think the policy for clearing the cache is too parser-specific to be hardcoded like that.

What I am afraid of, is that without a strategy to clean caches we will simply be unable to run extensive tests. For example, without this simple strategy I was unable to run the java parser on a few tens of thousand of files (it was either slowing down terribly or going out of memory), while with this simple change it completed in a reasonable time and reliably. 

I think that without this, we can forget to apply similar approaches in the parsers we build, or having anyway to copy and paste logic from one parser to another. This is why I think having some basic caching strategy "by default" could be useful.

I am wondering if we could get a "basic" cache-cleaning strategy in place and then let the possibility to override it in case parsers have special need. In this way we could combine the best of both worlds. Would this make sense?

We could also have a smarter basic cache-cleaning strategy. At the very least one could customize the number of iterations among the cache cleaning, but one could also trigger the cache cleaning based on other factors, like for example the memory occupied or the size of these caches.

Would this work?

alessiostalla:
Yes I had something like that in mind.

alessiostalla:
And of course all this reasoning only applies if one reuses the same `KolasuParser` instance instead of creating a new one each time, so I think we should document this as a best practice if it isn't already.

ftomassetti:
Thank you Alessio. I will just add an overridable method to contain the logic on when to clear the caches and then merge it

PR #264, opened 2023-09-28
ASTCodeGenerator: allow to customize the PrinterOutput.

### LINKED ISSUE #263: ASTCodeGenerator: allow to customize the PrinterOutput
To set properties such as indentation and newlines

alessiostalla:
Fixes #263

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
@@ -31,14 +31,15 @@ abstract class ASTCodeGenerator<R : Node> {
 
     fun printToString(ast: R): String {
         val printerOutput = PrinterOutput(this.nodePrinters, nodePrinterOverrider)
-        return customizePrinter(printerOutput).apply { this.print(ast, prefix, postfix) }.text()
+        configurePrinter(printerOutput)
+        return printerOutput.apply { this.print(ast, prefix, postfix) }.text()
     }
 
     fun printToFile(root: R, file: File) {
         file.writeText(printToString(root))
     }
 
-    protected open fun customizePrinter(printerOutput: PrinterOutput): PrinterOutput = printerOutput
+    protected open fun configurePrinter(printerOutput: PrinterOutput) {}
 
     open val prefix: String get() = ""
     open val postfix: String get() = ""
--- a/core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
@@ -5,13 +5,13 @@ import java.io.File
 import kotlin.reflect.KClass
 
 /**
- * Transform an AST into code. This can be done on an AST obtained from parsing, or build programmatically.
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
```diff
@@ -30,12 +30,15 @@ abstract class ASTCodeGenerator<R : Node> {
...
-            .apply { this.print(ast) }
-            .text()
+        return printerOutput().apply { this.print(ast, prefix, postfix) }.text()
     }
 
     fun printToFile(root: R, file: File) {
         file.writeText(printToString(root))
     }
+
+    open fun printerOutput() = PrinterOutput(this.nodePrinters, nodePrinterOverrider)
```
ftomassetti, line 40: Why should someone be able to override this method?

Would not be confusing to set up nodePrinters and then have this method overriden and the nodePrinters being ignored?

core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
```diff
@@ -30,12 +30,15 @@ abstract class ASTCodeGenerator<R : Node> {
...
+
+    open val prefix: String get() = ""
```
ftomassetti, line 42: Why would someone need prefix and postfix? Could you indicate some examples?

ftomassetti (commented):
Added comments

core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
alessiostalla: Because they may want to customize the printer output (changing indentation, newlines, etc.)
Indeed overriding per se is not the only way. We could add a method `customizePrinterOutput`.

core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
alessiostalla: Prefix and postfix are exposed as parameters by the printer output, but `ASTCodeGenerator`doesn't allow to pass them. Anyway for prefix I have a couple of use cases such as adding information at the top of the file (e.g. licensing, or notes about the file having been generated, etc.) and adding for example a shebang line for scripts (or for a client the `#! CLIENT_NAME_REDACTED-java` header)

core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
ftomassetti: Would that be a sort of post-processing? If so one could make a `postProcessing` method which by default does nothing and make that one overridable, otherwise I think your approach would be best

core/src/main/kotlin/com/strumenta/kolasu/codegen/ASTCodeGenerator.kt
ftomassetti: Oh, right, it makes sense. Can you add this as a comment?

ftomassetti:
+1 for me

Issue #267, opened 2023-10-16
PrinterOutput: unnecessary wrapping of exceptions in RuntimeException


alessiostalla:
Seems like a Javaism to me. Kotlin exceptions are not checked and don't require wrapping. Also it's catching Throwable which is an anti-pattern (swallows everything including system exceptions such as heap exhausted).

PR #272, opened 2023-10-24
PrinterOutput: don't wrap exceptions

alessiostalla:
This fixes https://github.com/Strumenta/kolasu/issues/267

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/codegen/PrinterOutput.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/codegen/PrinterOutput.kt
@@ -121,11 +121,7 @@ class PrinterOutput(
         print(prefix)
         val printer = getPrinter(ast)
         associate(ast) {
-            try {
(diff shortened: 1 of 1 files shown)


ftomassetti:
The additional information here is the AST that could not be printed. I think this is useful for debugging the printer

alessiostalla:
But you already know that the AST could not be printed due to `printToString` throwing an exception. This forces you to check on the exception's cause:

```kotlin
fun generateCode(root: ROOT, issues: MutableList<Issue>) = try {
    SomeCustomerCodeGenerator().printToString(root)
} catch (e: Exception) {
    val cause = e.cause
    if (cause is MemberNotResolvedException) {
        issues.add(Issue.translation(cause.message!!, position = cause.member.position))
    } else {
        issues.add(Issue.translation(e.message!!))
    }
    "// Code generation failed due to $cause, check reported issues"
}
```

ftomassetti:
Ok, I trust you. I am merging the PR

PR #276, opened 2023-11-07
Fix AST classes generation from lionweb language.

AtanasM98:
AST generation changes that check whether a class should be open and also fixes a bug when trying to refer to other generated class.


ftomassetti:
Thank you @AtanasM98 !

PR #278, opened 2023-11-21
References as containments

### LINKED ISSUE #277: References represented as containments in EMF

alessiostalla:
This fixes #277. All tests pass. However, before merging, I'd like to verify its impact on the RPG parser and at least another parser.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
@@ -24,26 +24,13 @@ annotation class Mapped(val path: String = "")
 /**
  * Factory that, given a tree node, will instantiate the corresponding transformed node.
  */
-class NodeFactory<Source, Output : Node> {
-    val constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> List<Output>
-    var children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf()
-    var finalizer: (Output) -> Unit = {}
-    var skipChildren: Boolean = false
+class NodeFactory<Source, Output : Node>(
+    val constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> List<Output>,
+    var children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf(),
+    var finalizer: (Output) -> Unit = {},
+    var skipChildren: Boolean = false,
     var childrenSetAtConstruction: Boolean = false
-
-    constructor(
-        constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> List<Output>,
-        children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf(),
-        finalizer: (Output) -> Unit = {},
-        skipChildren: Boolean = false,
-        childrenSetAtConstruction: Boolean = false
-    ) {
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -366,7 +353,7 @@ open class ASTTransformer(
         val childFactory = childNodeFactory as ChildNodeFactory<Any, Any, Any>
         val childrenSource = childFactory.get(getSource(node, source))
         val child: Any? = if (pd.multiple) {
-            (childrenSource as List<*>).map { transformIntoNodes(it, node) }.flatten()
+            (childrenSource as List<*>?)?.map { transformIntoNodes(it, node) }?.flatten()
```
ftomassetti, line 356: I think lists of children should never be null. Do we have a reason for allowing null lists?

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -24,26 +24,13 @@ annotation class Mapped(val path: String = "")
 /**
  * Factory that, given a tree node, will instantiate the corresponding transformed node.
  */
-class NodeFactory<Source, Output : Node> {
-    val constructor: (Source, ASTTransformer, NodeFactory<Source, Output>) -> List<Output>
-    var children: MutableMap<String, ChildNodeFactory<Source, *, *>?> = mutableMapOf()
-    var finalizer: (Output) -> Unit = {}
-    var skipChildren: Boolean = false
+class NodeFactory<Source, Output : Node>(
```
ftomassetti, line 27: Better!

emf/src/main/kotlin/com/strumenta/kolasu/emf/MetamodelBuilder.kt
```diff
@@ -229,7 +232,8 @@ class MetamodelBuilder(packageName: String, nsURI: String, nsPrefix: String, res
     private fun registerReference(
```
ftomassetti, line 232: Maybe this name is confusing. Perhaps we could call it `Link`, where `Link` is the generic thing that could either be a Containment or a Reference

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
alessiostalla: Sorry, this wasn't meant for this PR. However, yes there is a use case. I encountered it in the SAS parser. If I remember correctly, in case of some parse errors, ANTLR can return null collections, and the transformer would crash.
However, probably it's better that the transformer always returns a non-null list.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
alessiostalla: I think IntelliJ did all the work :D

emf/src/main/kotlin/com/strumenta/kolasu/emf/MetamodelBuilder.kt
alessiostalla: Yes, makes sense.

alessiostalla:
Closing as invalid. The issue turned out to be in Tylasu. Kolasu is representing `ReferenceByName` instances as children in EMF, which is fine.

Issue #280, opened 2023-12-16
Assess migration of core to Kotlin Multi Platform 


ftomassetti:
- [ ] Move serialization out of core so we can drop dependencies from JSON and XML libraries
- [ ] Move to https://github.com/badoo/Reaktive

ftomassetti:
See also #281

PR #283, opened 2023-12-18
Extracting components from core module

ftomassetti:
Work towards #280 because by extraction components we remove JVM-only dependencies and therefore we move toward making possible to compile the core module for Kotlin multiplatform


alessiostalla (approved):

ftomassetti:
Thank you Alessio!

PR #284, opened 2023-12-18
Introduce the INode interface

### LINKED ISSUE #279: Make Node an interface

ftomassetti:
This will make possible to make interfaces such as `INamed` extend it, and therefore remove the concept of "Node Like" element.
Fix #279


alessiostalla (approved):
Looks good, I only have two observations:
- The name `INode` – I don't like the C#-like idiom of prepending interfaces with "I", also an inode is already an FS concept (at least on Linux), couldn't we call it "NodeLike" or something like that?
- What will be the impact of porting this to other languages, that either don't have interfaces at all (Python) or have them in limited form (TypeScript, where they exist only at compile-time so you can't write `x instanceof SomeInterface`)? In other words, is this intended only as an implementation detail for Kolasu, or is it a concept that we should apply over StarLasu?

ftomassetti:
> Looks good, I only have two observations:
> 
> * The name `INode` – I don't like the C#-like idiom of prepending interfaces with "I", also an inode is already an FS concept (at least on Linux), couldn't we call it "NodeLike" or something like that?

I also do not like the C#-like idiom. I think the right think would be to call the interface `Node` and change the name of the implementing class, however that would mean having to update the parsers significantly, so I tried to avoid that.
We have already an annotation called `NodeLike`, so calling the interface `NodeLike` could generate confusion. Also the suffix `Like` is not that obvious.

What about:
* calling the interface `Node`
* calling the implementing class `StandardNode` or `BaseNode` or `SimpleNode`?


> * What will be the impact of porting this to other languages, that either don't have interfaces at all (Python) or have them in limited form (TypeScript, where they exist only at compile-time so you can't write `x instanceof SomeInterface`)? In other words, is this intended only as an implementation detail for Kolasu, or is it a concept that we should apply over StarLasu?

I think of it mostly as an implementation detail, due to how the Kotlin typesystem works

ftomassetti:
In the end I followed @alessiostalla 's advice and renamed `INode` into `NodeLike`

PR #292, opened 2024-01-17
Improving parsing issues reporting

ftomassetti:
We do two things:
* We capitalize issues
* For issues that specify an issue on a token we specify the position for the issue to cover the whole token and not just the starting point of the token. Right now we specify "flat positions" (start == end), which means we cannot show those positions in editors

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
@@ -246,11 +246,16 @@ fun Parser.injectErrorCollectorInParser(issues: MutableList<Issue>) {
             errorMessage: String?,
             p5: RecognitionException?
         ) {
+            val startPoint = Point(line, charPositionInLine)
+            var endPoint = startPoint
+            if (p1 is CommonToken) {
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -238,18 +239,23 @@ fun Parser.injectErrorCollectorInParser(issues: MutableList<Issue>) {
...
             line: Int,
             charPositionInLine: Int,
             errorMessage: String?,
-            p5: RecognitionException?
+            recognitionException: RecognitionException?
         ) {
+            val startPoint = Point(line, charPositionInLine)
+            var endPoint = startPoint
+            if (offendingSymbol is CommonToken) {
+                endPoint = startPoint.plus(offendingSymbol.text)
```
alessiostalla, line 252: We have the extension method `Token.endPoint` (and `Token.position` that calls it) that take into account the EOF token corner case, so I suggest we use it here.

alessiostalla (approved):
This makes a lot of sense and I'll open a similar issue in Tylasu as well since it's so often used for editors.

PR #296, opened 2024-01-29
Make some modules multiplatform

ftomassetti:
This PR goes in the direction of supporting using of Kolasu with ANTLR Kotlin, so that a single Kolasu-based, multi-platform parser can be eventually delivered. We are not there yet as many core functionalities are JVM only.

The approach is to extract from `core` a module containing the absolutely base stuff (called `ast`).
We also rename the `antlr` module in `antlr4j` and we create a module named `antlr4k` (in the future we could have modules for antlr5).

A ton of files changed have changes only in the imports, as we move stuff into `ast` to a different package.


core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
```diff
@@ -2,12 +2,20 @@ package com.strumenta.kolasu.model
```
alessiostalla, line 2: Why is `Node` still in the _model_ package and not in the _ast_ package, given that everything else was moved?

alessiostalla (approved):
This looks good

core/src/main/kotlin/com/strumenta/kolasu/model/Node.kt
ftomassetti: Originally I put everything in the ast module under the ast package. But now I realize that it does not make sense. In our contest ast and model are synonym, so I will rename the ast package into model

ftomassetti:
> I miss the reason why we changed a class from `PropertyDescription` to `FeatureDescription`, but everything seems alright.

It is to align the naming with LionWeb (and for internal consistency). In general Feature means either an attribute, a containment or a reference

loradd (approved):
Looks good to me! 

My doubts where answered in the discussion with Alessio and Gabriele.

ftomassetti:
Thank you all!

PR #297, opened 2024-01-31
Fixes to support RPG Parser and LionWeb

ftomassetti:
* We add support for the Char type when converting to LionWeb
* When converting to LionWeb, we treat each type which is not a descendant of Node as a new Primitive type (note that it will require serializers and deserializers when converting ASTs to or from LionWeb)
* We support also derived properties as there was a case when `name` was marked as derived and it ended up being ignored


alessiostalla (approved):
Looks good, as far as I understand of Lionweb

mctaverna (approved):

ftomassetti:
Thank you!

PR #298, opened 2024-02-01
We add methods to get properties excluding derived properties

gabriele-tomassetti:
Kolasu 1.5.39 started reporting as children even properties with annotation `derived`. Testing a parser, we found an issue with that approach in some cases. In this parser a derived property contains also children included in the main node.

In this example, `child5`, a derived node, contains also the children of `child2`. The reason is that in the language statements could be in the global scope and also in a subroutine. This node is a virtual subroutine that contains the statements in the global scope. This provides a simple way for users to handle all statements in a program, by looking at all subroutines.  
```
root 
|> child1
|> child2 
     |> child3
     |> child4
|> child5 (derived)
     |> child7
          |> child3
          |> child4 
```

If we want to allow this behavior, we need to modify the `children` property of `Node`. Otherwise, we get loops, for example in the `JsonGenerator` class. Or, we also get multiple time the same childrens.

This branch contains some changes to support this use case (derived nodes that contains some children of the parent node). We have to decide if we want to allow this use case and if so, if this is the best way to support it.

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -71,6 +71,10 @@ open class Node() : Origin, Destination, Serializable {
             throw RuntimeException("Issue while getting properties of node ${this::class.qualifiedName}", e)
         }
 
+    /**
+     * The properties of this AST nodes, including attributes, children, and references, but excluding derived
+     * properties.
+     */
     @property:Internal
     open val originalProperties: List<PropertyDescription>
         get() = try {
@@ -230,12 +234,8 @@ val <T : Any> KClass<T>.nodeProperties: Collection<KProperty1<T, *>>
         .toList()
 
 val <T : Any> KClass<T>.nodeOriginalProperties: Collection<KProperty1<T, *>>
-    get() = memberProperties.asSequence()
-        .filter { it.visibility == KVisibility.PUBLIC }
+    get() = nodeProperties
         .filter { it.findAnnotation<Derived>() == null }
-        .filter { it.findAnnotation<Internal>() == null }
-        .filter { it.findAnnotation<Link>() == null }
-        .toList()
 
 /**
  * @return all properties of this node that are considered AST properties.
--- a/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
@@ -163,12 +163,15 @@ open class Node() : Origin, Destination, Serializable {
         return "${this.nodeType}(${properties.joinToString(", ") { "${it.name}=${it.valueToString()}" }})"
     }
 
-    fun getChildren(containment: Containment): List<Node> {
-        return getChildren(containment.name)
+    fun getChildren(containment: Containment, includeDerived: Boolean = false): List<Node> {
+        return getChildren(containment.name, includeDerived)
     }
 
-    fun getChildren(name: String): List<Node> {
-        return when (val rawValue = properties.find { it.name == name }!!.value) {
+    fun getChildren(name: String, includeDerived: Boolean = false): List<Node> {
+        return when (
+            val rawValue = (if (includeDerived) properties else originalProperties)
+                .find { it.name == name }!!.value
+        ) {
             null -> {
                 emptyList()
             }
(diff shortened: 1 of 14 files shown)


core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -212,19 +220,35 @@ fun <N : Node> N.withOrigin(origin: Origin?): N {
...
+val <T : Any> Class<T>.nodeOriginalProperties: Collection<KProperty1<T, *>>
+    get() = this.kotlin.nodeOriginalProperties
 val <T : Any> KClass<T>.nodeProperties: Collection<KProperty1<T, *>>
     get() = memberProperties.asSequence()
         .filter { it.visibility == KVisibility.PUBLIC }
         .filter { it.findAnnotation<Internal>() == null }
         .filter { it.findAnnotation<Link>() == null }
         .toList()
 
+val <T : Any> KClass<T>.nodeOriginalProperties: Collection<KProperty1<T, *>>
```
ftomassetti, line 232: Couldn't this be expressed as a filter on `nodeProperties`?

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -71,6 +71,14 @@ open class Node() : Origin, Destination, Serializable {
             throw RuntimeException("Issue while getting properties of node ${this::class.qualifiedName}", e)
         }
 
+    @property:Internal
+    open val originalProperties: List<PropertyDescription>
```
ftomassetti, line 75: A comment would be useful

core/src/main/kotlin/com/strumenta/kolasu/traversing/Structurally.kt
```diff
@@ -96,7 +96,7 @@ fun Node.walkAncestors(): Sequence<Node> {
  */
 fun Node.walkChildren(): Sequence<Node> {
     return sequence {
-        this@walkChildren.properties.forEach { property ->
+        this@walkChildren.originalProperties.forEach { property ->
```
ftomassetti, line 99: maybe also `walkChildren` could get the flag `includeDerived` set to false by default

javalib/src/test/java/com/strumenta/kolasu/javalib/TraversingTest.java
```diff
@@ -2,6 +2,7 @@
 
 import com.strumenta.kolasu.model.Node;
 import com.strumenta.kolasu.model.Processing;
+import com.strumenta.kolasu.model.PropertyDescription;
```
ftomassetti, line 5: Why?

ftomassetti:
In general it looks good, I just added a few comments

javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
```diff
@@ -22,6 +22,13 @@ public List<B> getBs() {
         public List<PropertyDescription> getProperties() {
             return Arrays.asList(new PropertyDescription("bs", true, Multiplicity.MANY, getBs(), PropertyType.CONTAINMENT, false));
         }
+
+        @NotNull
+        @Override
+        @Internal
+        public List<PropertyDescription> getOriginalProperties() {
```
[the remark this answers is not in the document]
gabriele-tomassetti: I addressed other comments, but I am not exactly sure what you mean by this. I changed the method to return `getProperties`.

core/src/main/kotlin/com/strumenta/kolasu/model/Model.kt
```diff
@@ -212,19 +220,35 @@ fun <N : Node> N.withOrigin(origin: Origin?): N {
...
+    get() = this.kotlin.nodeOriginalProperties
 val <T : Any> KClass<T>.nodeProperties: Collection<KProperty1<T, *>>
     get() = memberProperties.asSequence()
         .filter { it.visibility == KVisibility.PUBLIC }
         .filter { it.findAnnotation<Internal>() == null }
         .filter { it.findAnnotation<Link>() == null }
         .toList()
 
+val <T : Any> KClass<T>.nodeOriginalProperties: Collection<KProperty1<T, *>>
+    get() = memberProperties.asSequence()
```
ftomassetti:
```suggestion
    get() = nodeProperties()
        .filter { it.findAnnotation<Derived>() == null }
```

javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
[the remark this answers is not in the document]
ftomassetti: If `getOriginalProperties` get implemented as `getProperties().filter { /* if not derived */ }` then you do not need to override it here

gabriele-tomassetti:
Ready to merge. There is a linting issue, but given the force push I prefer to not touch it.

ftomassetti:
Thank you @gabriele-tomassetti

PR #299, opened 2024-02-09
Avoid conflict in lion web conversion

ftomassetti:
Nodes which are "equals" but are not the same identity, cause issues when converting to LionWeb.


ftomassetti:
With these fixes I can serialize several Java ASTs to LionWeb

alessiostalla:
Looks good, I don't quite understand which problem it's solving (i.e. which AST patterns were failing) so maybe you could add a test exemplifying the problem?

alessiostalla (approved):

ftomassetti:
Some tests covers the root cause: the fact is that certain Kolasu Nodes which are "equals" are not the same node for LionWeb, so we need to change some operations to use identity and not equals

ftomassetti:
Thank you @alessiostalla !

PR #300, opened 2024-02-15
Introducing the LionWebNodeIdProvider

ftomassetti:
We introduce an interface defining the logic used to produce LionWeb Node IDs for Kolasu Nodes.
This logic could be used to "predict" which ID will be associated to each node. We will also be able to recalculate that ID, if we have access to the instance of the LionWebNodeIdProvider. 

This is relevant for the Code Insight Studio and for the symbol resolution.


loradd:
As part of the implementation of the symbol resolution service, I have defined a generic identifier provider interface. A default declarative subclass is provided to define providers based on node types, i.e. `identifierFor(nodeType: KClass<out Node>): String?`. 

The approach described in this PR seems compatible to me, we might want to review it all once everything will be integrated.

alessiostalla (approved):
For the Code Insight Studio, this looks good!

ftomassetti:
Thank you Alessio!

PR #301, opened 2024-02-20
Semantic Enrichment Support

loradd:
This PR introduces a `semantics` module containing all components that can currently be used to perform semantic enrichment activities, e.g. symbol resolution.

At the moment, the module also contains a `NodeRepository` interface, which I suggest we could move into a separate `persistence` module or `com.strumenta.kolasu.persistence` package in `core`. Repository or persistence for this suggestion is completely temporary? The general idea is that it represents something that we use to store nodes, e.g. from an in-memory map to a LionWeb repository.


loradd:
Unit tests and proper documentation missing at the moment, will be done after our internal experiments with the Java langmodule.

loradd:
I've added some deprecation annotations on old components in `kolasu-core`

ftomassetti:
Before merging, I would perform two verifications. To perform these two verifications I would build a local version of Kolasu with these changes.
The verifications are:
1. We can build a simple semantics module (for example for Java)
2. Have some functional tests in https://github.com/Strumenta/starlasu-lionweb-repository-client perform symbol resolution, store the resolved AST and retrieving it

I am reflecting on the latter point, to see if there is any conceptual challenge regarding the integration with the lionweb-repository, so that we can discuss it next week

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/identifier/provider/IdentifierProvider.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.strumenta.kolasu.model.Node
+import kotlin.reflect.KClass
+
+interface IdentifierProvider {
+    /**
+     * Retrieve the identifier associated
+     * with the given node (possibly cast as requested)
+     **/
+    fun <NodeTy : Node> getIdentifierFor(
+        node: NodeTy,
```
ftomassetti, line 12: I would add a comment to explain why someone may want to pass `typedAs`

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,39 @@
+package com.strumenta.kolasu.semantics.symbol.description
+
+data class SymbolDescription(
```
ftomassetti, line 3: I think we should add a comment to indicate that they can represent nodes or builtin elements. And the fact that they represent a "slim" version of nodes, to be used in the Symbol Resolution Index

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,39 @@
...
+
+data class ReferenceValueDescription(
+    override val value: SymbolDescription? = null
+) : ValueDescription
+
+data class ListValueDescription(
+    override val value: List<ValueDescription> = emptyList()
+) : ValueDescription
+
+data class SetValueDescription(
```
ftomassetti, line 33: We do not use set in Kolasu

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,39 @@
+package com.strumenta.kolasu.semantics.symbol.description
+
+data class SymbolDescription(
+    val identifier: String,
+    val types: List<String>,
+    val properties: Map<String, ValueDescription>
+)
+
+sealed interface ValueDescription {
```
ftomassetti, line 9: We may also want to use enums or arbitrary primitive types

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,39 @@
...
+
+data class ListValueDescription(
+    override val value: List<ValueDescription> = emptyList()
+) : ValueDescription
+
+data class SetValueDescription(
+    override val value: Set<ValueDescription> = emptySet()
+) : ValueDescription
+
+data class MapValueDescription(
```
ftomassetti, line 37: We do not use maps in Kolasu

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: Perhaps we could keep it for future developments?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: `StringValueDescription`, `IntegerValueDescription` and `BooleanValueDescription` currently support the specification of primitive typed values. These are just wrappers to avoid using a map of `Any` typed properties and maintain control over which kinds of values we handle.

Enumerations can be added fairly easy - I will keep this comment as remainder before merging the PR.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: I agree, the current implementation is not really well documented as the main focus consisted in experimenting first. I will add a detailed explanation of every component once done, thanks.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/identifier/provider/IdentifierProvider.kt
loradd: I agree, the current implementation is not really well documented as the main focus consisted in experimenting first. I will add a detailed explanation of every component once done, thanks.

loradd:
> Before merging, I would perform two verifications. To perform these two verifications I would build a local version of Kolasu with these changes. The verifications are:
> 
> 1. We can build a simple semantics module (for example for Java)
> 2. Have some functional tests in https://github.com/Strumenta/starlasu-lionweb-repository-client perform symbol resolution, store the resolved AST and retrieving it
> 
> I am reflecting on the latter point, to see if there is any conceptual challenge regarding the integration with the lionweb-repository, so that we can discuss it next week

I agree with this plan, as discussed before. This PR is meant to start discussing the integration of this module. I will convert it to draft.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: I would rather not because there is a lot of code that depends on the assumption that we restrict what goes into Nodes (and I see Symbols as a "view" of nodes).

For example: transformers, Json serialization, Xml serialization, export to LionWeb, starlasu tools, IDEA plugin, compiler plugins, etc.

Supporting more types in all these aspects would be a lot of work and it would make the codebase significantly more complex. It would means also making the library more difficult to learn and to extend.

Supporting Sets or Maps in only a subset of the functionalities of Kolasu, could lead to errors while using the library

loradd: I agree, I will remove sets and maps value descriptions then

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: Set and Map value descriptions will be removed (see discussion above)

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: Thank you, I will write down in the wiki a page about what can go in the AST and what cannot go in it, and the reasons for it because I realize I never actually wrote about thi

alessiostalla:
I'd wait for an example application to a language module, too. Also @loradd did you look at [GitHub's stack graphs](https://github.blog/2021-12-09-introducing-stack-graphs/)? There are a few talks around. I like the fact that they abstract the symbols and the resolution rules as graph traversal operations that are incremental. Especially when the resolution depends on a "middle man" (such as a module definition importing some symbols and reexporting them, maybe even under a different name), they can recompute the references incrementally when just the "middle man" changes even if neither the AST using the reference, nor the AST where the reference (logically) points to, have changed.

loradd:
> I'd wait for an example application to a language module, too. 

It makes sense, the plan is to start using this approach in the Java language module. In the meantime, I thought about opening this PR to start discussing the various technical decisions I took while developing the various components.

> Also @loradd did you look at [GitHub's stack graphs](https://github.blog/2021-12-09-introducing-stack-graphs/)? There are a few talks around. I like the fact that they abstract the symbols and the resolution rules as graph traversal operations that are incremental. Especially when the resolution depends on a "middle man" (such as a module definition importing some symbols and reexporting them, maybe even under a different name), they can recompute the references incrementally when just the "middle man" changes even if neither the AST using the reference, nor the AST where the reference (logically) points to, have changed.

Yes, I also had an interesting conversation with Aaron Zwan at LangDev about these - he presented Scope Graphs, which are what Stack Graphs are implementing more or less. In our approach, we also have a generic representation for symbols - SymbolDescription - and introducing a reactive update mechanism would be possible. Indeed, referencing and referenced nodes can be identified by inspecting the symbols in the index. However, this was left as future development given the current time constraints.

loradd:
I realize now that there is no `ValueDescription` to represent `Node` containments - therefore, I will soon add a `ContainmentValueDescription` to this. Then, `ReferenceValueDescription` will represent simple references, whereas `ContainmentValueDescription` will represent actual containments.

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
```diff
@@ -34,8 +34,11 @@ interface Named : PossiblyNamed {
  * This is not statically enforced as we may want to use some interface, which cannot extend Node.
  * However, this is enforced dynamically.
  */
-class ReferenceByName<N>(val name: String, initialReferred: N? = null) : Serializable where N : PossiblyNamed {
-
+class ReferenceByName<N : PossiblyNamed>(
```
ftomassetti, line 37: I would add this in the KDoc.

```
ReferenceByName can indicate a reference to a Node present in the same AST or coming from external sources (typically other ASTs). In the latter case there are situations where we want to indicate that we know which Nodes the ReferenceByName refers to, but we do not want to retrieve it straight away for performance reason. In these circumstances the `referred` field will be null, but we will use an `identifier` field to indicate the referred Node that we may want to retrieve at a later stage.

So a ReferenceByName can be in three states:
* unresolved (referred and identifier set to null)
* resolved and retrieved (referred is not null, irrespectively of identifier being null or not)
* resolved but not retrieved (identifier is not null, referred is null)
```

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
```diff
@@ -49,6 +52,9 @@ class ReferenceByName<N>(val name: String, initialReferred: N? = null) : Seriali
         this.referred = initialReferred
     }
 
+    val resolved: Boolean
+        get() = referred != null
```
ftomassetti, line 56: Here I would say that the reference is resolved when either referred or identified are different from null. To me resolved equals false indicates that we cannot find a Node to associate to this reference, while in some cases we can find it, but we do not retrieve it straight away. The reference is still resolved, from my point of view.

I would consider adding another property called `resolvedAndRetrieved` and I would make that one checking just `referred`

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
```diff
@@ -57,20 +63,15 @@ class ReferenceByName<N>(val name: String, initialReferred: N? = null) : Seriali
...
     }
 
-    val resolved: Boolean
-        get() = referred != null
-
     override fun hashCode(): Int {
         return name.hashCode() * (7 + if (resolved) 2 else 1)
     }
 
     override fun equals(other: Any?): Boolean {
```
ftomassetti, line 70: We should consider also `identifie` in `equals` (and possibly in `hashCode` too)

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/identifier/provider/declarative/DeclarativeIdentifierProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+        node: NodeTy,
+        typedAs: KClass<in NodeTy>?
+    ): String? {
+        return this.rules
+            .firstOrNull { it.canBeInvokedWith(typedAs ?: node::class) }
+            ?.invoke(this, node)
+    }
+}
+
+class DeclarativeIdentifierProviderRule<NodeTy : Node>(
```
ftomassetti, line 30: What is the role of this class?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/identifier/provider/declarative/DeclarativeIdentifierProvider.kt
```diff
@@ -0,0 +1,60 @@
...
+    override fun compareTo(other: DeclarativeIdentifierProviderRule<*>): Int {
+        return when {
+            this.nodeType.isSuperclassOf(other.nodeType) -> 1
+            this.nodeType.isSubclassOf(other.nodeType) -> -1
+            else -> (this.nodeType.qualifiedName ?: "") compareTo (other.nodeType.qualifiedName ?: "")
+        }
+    }
+}
+
+data class DeclarativeIdentifierProviderRuleArguments<NodeTy : Node>(
```
ftomassetti, line 57: What is the role of this class?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/node/repository/NodeRepository.kt
```diff
@@ -0,0 +1,9 @@
+package com.strumenta.kolasu.semantics.node.repository
+
+import com.strumenta.kolasu.model.Node
+
+interface NodeRepository {
+    fun load(identifier: String): Node?
+
+    fun store(node: Node): String?
```
ftomassetti, line 8: When do we need to store nodes?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/importer/SymbolImporter.kt
```diff
@@ -0,0 +1,20 @@
+package com.strumenta.kolasu.semantics.symbol.importer
+
+import com.strumenta.kolasu.model.Node
+import com.strumenta.kolasu.model.children
+import com.strumenta.kolasu.semantics.symbol.provider.SymbolProvider
+import com.strumenta.kolasu.semantics.symbol.repository.SymbolRepository
+
+open class SymbolImporter(
```
ftomassetti, line 8: What is the role of a SymbolImporter? Does it translate nodes into SymbolDescriptions?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
```diff
@@ -0,0 +1,146 @@
...
+
+    private fun toValueDescription(identifierProvider: IdentifierProvider, source: Any?): ValueDescription? {
+        return when (source) {
+            is Boolean -> BooleanValueDescription(source)
+            is Int -> IntegerValueDescription(source)
+            is String -> StringValueDescription(source)
+            is Node -> toContainmentValueDescription(identifierProvider, source)
+            is ReferenceByName<*> -> toReferenceValueDescription(identifierProvider, source)
+            is List<*> -> toListValueDescription(identifierProvider, source)
+            else -> null
```
ftomassetti, line 110: What about Enum? What about Char? What about arbitrary primitive types? Are they excluded on purpose?

core/src/main/kotlin/com/strumenta/kolasu/model/Naming.kt
loradd: Done

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/identifier/provider/declarative/DeclarativeIdentifierProvider.kt
loradd: This is now renamed to `DeclarativeIdentifierProviderRuleContext` and it contains everything that can be accessed while defining an identifier provider rule.
```kotlin
object MyIdentifierProvider: DeclarativeIdentifierProvider(
    identifierFor(SomeNode::class) {
        // if nothing is specified, the context will be accessible through the it variable    
    },
    identifierFor(SomeNode::class) { (node: SomeNode, identifiers: IdentifierProvider) ->
        // node is the current node we are defining the identifier for
        // identifierProvider can be used to compute identifiers for other nodes
        // It could be used to compute the identifier of the container node, for example.
    }
)
```
This mechanism allows us to reuse the identifier inside its configuration, we would not be able to use `MyIdentifierProvider` otherwise - at least not using `val` or `object` definitions. We would probably need to subclass `DeclarativeIdentifierProvider` and this is considered as anti-pattern of our symbol resolution APIs.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/node/repository/NodeRepository.kt
loradd: Good question! :) Now that the symbol resolution does not need to retrieve nodes for global symbols, we never need access to a node repository - it was removed in the last commit.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/importer/SymbolImporter.kt
loradd: Exactly, the `SymbolImporter` can be used to transform an AST into its corresponding `SymbolDescription` instances and store these into the `SymbolRepository` instance.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
loradd: Not excluded on purpose, just a matter of time constraints. We can gradually introduce new types while experimenting with the APIs.

ftomassetti:
@loradd I am using the combination of this and #302 to test locally the functional test involving symbol resolution in [branch symres of the starlasu-lionweb-repository-client](https://github.com/Strumenta/starlasu-lionweb-repository-client/blob/6300862188e09ceca31289b77056e5865cac896f/starlasu-client/src/functionalTest/kotlin/TodoFunctionalTest.kt#L117)

Once we have all the changes needed to made that test pass in here, I think we could go ahead and merge this one, making a release of Kolasu and celebrate.

In my opinion what should we do is:
1. Rebase this PR on master again (as I merged #305 to remove some spurious changes from #302 and make it easier to review it)
2. Rebase #302 on #301
3. You could then review #302 (I expect you would less ~10 files changed and not 15 like there are now)
4. We could merge #302 in this
5. We could verify that the functional test involving symbol resolution still pass
6. We could merge this PR
7. We could release Kolasu
8. We could verify the functional test involving resolution pass using the official Kolasu release

What do you think?
Would this work for you?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,46 @@
+package com.strumenta.kolasu.semantics.symbol.description
+
+import com.strumenta.kolasu.language.KolasuLanguage
+import com.strumenta.kolasu.model.Named
+import com.strumenta.kolasu.model.Node
+
+data class SymbolDescription(
+    override val name: String,
```
ftomassetti, line 8: What happens if I specify a field called "name"? Would this overlap with the name field or be separated?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
ftomassetti: For values that are currently not supported, should we throw an exception instead of returning null?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/resolver/SymbolResolver.kt
```diff
@@ -0,0 +1,70 @@
...
+            ?.let { this.scopeProvider.scopeFor(node, reference)?.resolve(it) }
+    }
+
+    /**
+     * Attempts to resolve all reference properties of the
+     * given node and its children (if `withChildren` is `true`).
+     **/
+    fun resolve(
+        node: Node,
+        withChildren: Boolean = false
```
ftomassetti, line 53: This seems to resolve only the immediate children in #302 I changed the flag to run resolve on the whole tree (1. I renamed the flag 2. I passed the flag to the call to `resolve` on the children)

loradd: Ahhh, it's a bug - it should be `this.resolve(it, withChildren)` instead. I am fine with renaming `withChildren` with `entireTree`, I will apply this change on this branch as well.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: The two fields would be separated - we could handle the alignment in the DeclarativeSymbolProvider API. For example, whenever including a property named `name`, we could also se the name of the symbol. At the moment, the name can be set as follows:
```kotlin
object MySymbolProvider: DeclarativeSymbolProvider(
    symbolFor(SomeNode::class) {
        name("some name") // <- this sets the name of the symbol description
        include(SomeNode::name) // <- we could also implicitly set the name when including properties like these
    }
)
```
The `name()` directive could be removed, but the symbol description name would be null if no `name` property is included. What would happen with cases where a name property is not defined but we still want to define a symbol name. Perhaps we can keep both and specify that the last writer wins in the documentation. In the example above, the final name value will be given by `SomeNode::name`. What do you think?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
loradd: Okay, I will replace null with a RuntimeException

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
```diff
@@ -0,0 +1,165 @@
...
+    private val nodeIdProvider: NodeIdProvider,
+    vararg rules: DeclarativeSymbolProviderRule<out Node>
+) : SymbolProvider {
+    private val rules: List<DeclarativeSymbolProviderRule<out Node>> = rules.sorted()
+
+    override fun symbolFor(node: Node): SymbolDescription {
+        return this.rules
+            .firstOrNull { it.isCompatibleWith(node::class) }
+            ?.invoke(nodeIdProvider, this, node)
+            ?: throw RuntimeException("No compatible rule found for ${node::class.qualifiedName}")
```
ftomassetti, line 40: What is not clear to me is what I should do to declare symbols only for some nodes. Before I think I was just not providing a rule in the DeclarativeSymbolProvider, but now it seems that I must always provide a symbol for a node. I thought that the idea was to provide symbols only for nodes that we want to make visibile from outside the AST

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+ * Can be used whenever listing the rules in the DeclarativeNodeIdProvider constructor.
+ **/
+inline fun <reified NodeTy : Node> idFor(
+    noinline specification: (DeclarativeNodeIdProviderRuleContext<NodeTy>) -> String
+): DeclarativeNodeIdProviderRule<NodeTy> = DeclarativeNodeIdProviderRule(NodeTy::class, specification)
+
+/**
+ * Class representing a single rule of a DeclarativeNodeIdProvider.
+ **/
+class DeclarativeNodeIdProviderRule<NodeTy : Node>(
```
ftomassetti, line 44: If this class is never instantiated directly but always through `idFor`, could it make sense to make it private?

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlin.reflect.full.isSuperclassOf
+
+/**
+ * Declarative node id provider instances allow to define
+ * language-specific or custom identifier construction policies.
+ *
+ * ```kotlin
+ * object MyDeclarativeNodeIdProvider: DeclarativeNodeIdProvider(
+ *     idFor(PackageDeclaration::class) { it.node.name }
+ *     idFor(ClassDeclaration::class) { (node: ClassDeclaration, identifiers: NodeIdProvider) ->
```
ftomassetti, line 15: Is the second parameter of the lambda (`identifies`) always corresponding the the instance of the NodeIdProvider being defined (`MyDeclarativeNodeIdProvider` in this case)?
If so, can we add it through `this`?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -0,0 +1,134 @@
...
+ * - try to retrieve a local symbol associated with the given reference name;
+ * - if not found, try to retrieve a global symbol associated with the given reference name;
+ * - if not found, delegate the resolution to its parent (if any)
+ *
+ **/
+class ScopeDescription(
+    private val ignoreCase: Boolean = false
+) : ScopeDescriptionApi {
+    private var parent: ScopeDescription? = null
+    private val identifiers: MutableMap<String, String> = mutableMapOf()
```
ftomassetti, line 41: Is identifiers a map from name to identifiers? If so I suggest renaming it `namedToIdentifiers`, because it would be easy to invert name and identifier when using it

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -0,0 +1,134 @@
...
+    private val nodes: MutableMap<String, PossiblyNamed> = mutableMapOf()
```
ftomassetti, line 42: Is this a map from names to Symbols? If so, I suggest renaming it `namesToSymbols` to make it more obvious. Otherwise one could perhaps pass identifiers to it

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/provider/declarative/DeclarativeScopeProvider.kt
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    fun canBeInvokedWith(
+        nodeType: KClass<*>,
+        property: KProperty1<*, ReferenceByName<out PossiblyNamed>?>
+    ): Boolean {
+        return this.nodeType.isSuperclassOf(nodeType) && this.propertyName == property.name
+    }
+
+    override fun compareTo(other: DeclarativeScopeProviderRule<out Node>): Int {
```
ftomassetti, line 92: This seems logic we have used elsewhere. Would it make sense to create a utility method called "orderBySpecificity"?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -0,0 +1,50 @@
...
+    val identifier: String,
+    val types: List<String>,
+    val fields: Map<String, ValueDescription>
+) : Node(), Named
```
ftomassetti, line 12: When working on serializing symbols to LionWeb I realized it was not so much work to do so, so we do not need to make symbols into nodes for serialization purposes. Without that reason, do we still it still makes sense to have symbols as nodes?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/SymbolProvider.kt
```diff
@@ -0,0 +1,8 @@
+package com.strumenta.kolasu.semantics.symbol.provider
+
+import com.strumenta.kolasu.model.Node
+import com.strumenta.kolasu.semantics.symbol.description.SymbolDescription
+
+interface SymbolProvider {
+    fun symbolFor(node: Node): SymbolDescription
```
ftomassetti, line 7: I would expect the symbol provider _not_ to produce symbols for all nodes that are not visible outside the AST

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
```diff
@@ -0,0 +1,165 @@
...
+    private val specification: DeclarativeSymbolProvideRuleApi<NodeTy>.(
+        DeclarativeSymbolProviderRuleContext<NodeTy>
+    ) -> Unit
+) : DeclarativeSymbolProvideRuleApi<NodeTy>,
+    (NodeIdProvider, SymbolProvider, Node) -> SymbolDescription,
+    Comparable<DeclarativeSymbolProviderRule<out Node>> {
+    private var name: (Node) -> String? = { null }
+    private val properties: MutableMap<String, (NodeIdProvider, Node) -> ValueDescription> = mutableMapOf()
+
+    override fun name(name: String) {
```
ftomassetti, line 55: If every symbol must always have a name, should we get the name in the constructor? Or verify somehow that the name has been provided?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
loradd: You're right, I will restore the previous interface.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: It would produce an error as the `DeclarativeNodeIdProvider` takes instances of this class as input. The `idFor` is just an inlined utility method to build instances of these, which therefore need  to be publicly accessible.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: Yes, we could move the node id provider as `this`, move `node` as `it` and remove the context class. Do you something like this?
```kotlin
NodeIdProvider.(NodeTy) -> String
```

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
loradd: I agree, will do it

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
loradd: I agree, will do it

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/provider/declarative/DeclarativeScopeProvider.kt
loradd: There is a similar extension on KClass<*> in `com.strumenta.kolasu.utils` - this works on KType. I could move it there.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/SymbolProvider.kt
loradd: Correct, will fix it

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/resolver/SymbolResolver.kt
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Symbol resolver instances can be used to resolve references within AST nodes
+ * (i.e. `ReferenceByName` instances). Internally, the symbol resolver uses a scope provider
+ * defining the language-specific scoping rules. The resolution process can be invoked for
+ * specific node properties or for all properties of a given node (possibly its children as well).
+ *
+ * Symbol resolution can be executed invoking one of the following methods:
+ * - `resolve(node, reference)`: to resolve a specific reference of the given node;
+ * - `resolve(node, withChildren)`: to resolve all references of the given node
```
ftomassetti, line 21: This parameter is now called `entireTree`

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/provider/declarative/DeclarativeSymbolProvider.kt
ftomassetti: Makes sense

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/importer/SymbolImporter.kt
ftomassetti: It could be useful to add a very simple comment to explain this

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -0,0 +1,134 @@
...
+        symbol: Node
+    ) {
+        when (symbol) {
+            is SymbolDescription -> this.namesToExternalSymbolIdentifiers[name.asKey()] = symbol.identifier
+            is PossiblyNamed -> this.namesToLocalSymbolNodes[name.asKey()] = symbol
+        }
+    }
+
+    override fun define(symbol: PossiblyNamed) {
+        if (symbol is Node && symbol.name != null) {
```
ftomassetti, line 76: What happens if a pass something that is not a node? Should an exception be thrown?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -0,0 +1,134 @@
...
+ * - try to retrieve a local symbol associated with the given reference name;
+ * - if not found, try to retrieve a global symbol associated with the given reference name;
+ * - if not found, delegate the resolution to its parent (if any)
+ *
+ **/
+class ScopeDescription(
+    private val ignoreCase: Boolean = false
+) : ScopeDescriptionApi {
+    private var parent: ScopeDescription? = null
+    private val namesToExternalSymbolIdentifiers: MutableMap<String, String> = mutableMapOf()
```
ftomassetti, line 41: Good names!

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: No, I was just thinking that:

```
object MyDeclarativeNodeIdProvider: DeclarativeNodeIdProvider(
      idFor(PackageDeclaration::class) { /* here I can see the NodeIdProvider using this, I do not need to receive it as parameter of the lambda */ }
}
```

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: Then could we make its constructor private or internal, to communicate that people should not use it directly?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
loradd: Right, I will add the following exception:
```kotlin
RuntimeException("Symbols must be SymbolDescription or Node instances with a non-null name property."
```

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/resolver/SymbolResolver.kt
loradd: Ops, sorry, you're right - will fix it.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: Exactly, this is now implemented :)

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -0,0 +1,134 @@
...
+    ignoreCase: Boolean = false,
+    init: ScopeDescriptionApi.() -> Unit
+): ScopeDescription = ScopeDescription(ignoreCase).apply(init)
+
+/**
+ * A scope description containing all relevant information
+ * concerning the visible elements at a given point in space or time.
+ * - `ignoreCase`: whether the names associated to the symbols
+ * in this scope should be handled as case-sensitive or not;
+ * - `nodes`: the local symbols contained in this scope and
```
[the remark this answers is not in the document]
loradd: Ops, sorry, you're right - will fix it.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: It would still produce an error as `idFor` is inlined. This function is just syntactic sugar, users could actually use the `DeclarativeNodeIdProviderRule` constructor if they want. We would not be exposing internals anyway.

ftomassetti: Ok, marking this as resolved

ftomassetti:
@loradd I think we can now close the remaining comments, and merge this one.
Or do you think there is anything else left to do?

loradd:
> @loradd I think we can now close the remaining comments, and merge this one. Or do you think there is anything else left to do?

Yes, I think we can merge this one. I would like to review the API using @DslMarker annotations in the future, but I could open a separate issue for that, what do you think?

ftomassetti:
Perfect, thank you @loradd . Yes, we can do the other changes as follow up task. This is a quite nice achievement for this library!

PR #302, opened 2024-02-21
Experiments with SRI, solving references to symbols and not to nodes

ftomassetti:
This PR contains a set of changes done to make the [first functional test involving symbol resolution](https://github.com/Strumenta/starlasu-lionweb-repository-client/blob/48c36aad9f6b0b5ba394e00f0f97119258e1db18/starlasu-client/src/functionalTest/kotlin/TodoFunctionalTest.kt#L120) to work.

This PR depends on a PR in LionWeb Java (https://github.com/LionWeb-io/lionweb-java/pull/135)

As part of this PR I also:
* changed LionWebNodeIdProvider into NodeIdProvider and moved it to core, so that we can refer to it in the semantics module
* added a few tests in SymbolResolutionWithSRITest. They have been useful to play with the API and check how we could use it in a scenario close to the one we would get when using the Model Repository

Todo:
- [x] Get https://github.com/LionWeb-io/lionweb-java/pull/135 merged, release LionWeb Java and update it here
- [x] Separate the changes related to NodeIdProvider in a separate PR (-> #305 )
- [x] Have #301 rebase again on master, and then rebase this one on #301 to remove spurious changes from #305


loradd:
I am not sure I understand. Why do we want to use symbol descriptions in reference by names? It would not allow navigating the AST as `someNode.reference.referred?.somePropertyOfTheTarget`. The current implementation does not require external ASTs to be loaded, the following scenario could cover symbol resolution without an external model repository:

* DeclarativeSymbolProvider instance - defines what others can see from an other AST;
* DeclarativeScopeProvider instance - defines the scopes (name to identifier pairs) for each reference we want to resolve;
* SymbolImporter instance - takes every AST we need (even just one if external refs are not required) and populates the index using a SymbolProvider and SymbolRepository;
* SymbolResolver instance - takes the ASTs and resolves all references using a ScopeProvider and NodeRepository;
* SymbolRepository instance - stores the index somehow - might be an in-memory map or anything else;
* NodeRepository instance - stores the nodes somehow - might be in-memory or anything else;

Am I missing the point?

ftomassetti:
Thank you for your answer. This is a complex topic and it is very possible that we are not understanding each other and that I am missing something. I am happy to discuss about it when you have time.

In the meantime I share my understanding.

> I am not sure I understand. Why do we want to use symbol descriptions in reference by names? It would not allow navigating the AST as someNode.reference.referred?.somePropertyOfTheTarget. 

Right, that would be indeed a problem and a breaking change. In this case I would imagine that we could do this:
- Change the getter of `ReferenceByName.referred` to throw an exception if the symbol is set but not the node. The exception would instruct to load the node associated to the symbol before accessing it
- Add a method `ReferenceByName.loadNodeAssociatedToSymbol(nodeProvider)` to actually set the `referred` field to the real node pointed

> The current implementation does not require external ASTs to be loaded, 

I think it does, because we eventully get the nodes corresponding to the symbols and store those in references:

```
   // in SymbolResolver
    fun resolve(
        node: Node,
        property: KProperty1<Node, ReferenceByName<PossiblyNamed>>
    ) {
        @Suppress("UNCHECKED_CAST")
        val propertyValue: ReferenceByName<PossiblyNamed>? =
            node.properties
                .find { it.name == property.name }
                ?.value as? ReferenceByName<PossiblyNamed>?
        propertyValue
            ?.let { this.scopeProvider.scopeFor(node, property)?.resolve(it.name) }
            ?.let { this.nodeRepository.load(it) } 
            ?.let { propertyValue.referred = it as? PossiblyNamed }
    }
```

The actual node would be loaded when invoking `this.nodeRepository.load(it)`. Loading a node always means to get the entire sub-tree below it. For example, if we have a symbol pointing to a class declaration, this could would load the entire class declaration. Loading the class declaration we may find other references somewhere in that tree. If references can only hold nodes and not just symbols it means that loading those references we would end up loading more nodes (and their entire subtrees), and so on. Potentially we could end up loading from the model repository into memory a chain of ASTs.

I think that if we could load single nodes or proxies and not entire nodes we could avoid this issue, but currently I think we cannot.

My understanding is this: that `nodeRepository.load(symbol)` would in the implementation of interest get the data from the Model Repository and return the node. For example, let's suppose that we want to retrieve a `SumExpr`. 

```
data class SumExpr(val left: Expr, val right: Expr) : Node()
```

To instantiate `SumExpr` we would need to call the constructor, which requires passing its children (`left` and `right`) and so on. We cannot create a class `NodeProxy` to use instead of `Expr`, because that would not be compatible.

So for me the call `this.nodeRepository.load(it)` means load the entire sub-tree, as I would not know how to implement it differently.

Does this make sense? Am I missing a crucial point?

loradd:
You're right, loading the node would require loading its children as well. However, I think this is more an issue of the current `ReferenceByName` definition. Perhaps we could think of a way to re-design references independently of symbols. In general, I am not sure I like the idea of moving symbol descriptions into core as these should be considered specific to the semantic enrichment process.

We could rethink the current implementation of reference by names as follows:
* Reference by names keep track of a name and an identifier - provided by the NodeIdProvider of the language specification;
* Whenever asked, the actual referred node is loaded from the Node repository - all its reference by names will only contain the identifier, not the actual node;

The current `ReferenceByName` class specification is currently - more or less - this:
```kotlin
data class ReferenceByName<T: PossiblyNamed>(val name: String, val referred: T? = null)
```
we could change it as
```kotlin
class ReferenceTarget<T>(
    val identifier: String,
    val node: T? = null,
)

class ReferenceByName<T: PossiblyNamed>(
    val name: String,
    val target: ReferenceTarget<T>? = null,
)
```
When retrieved from the node repository `ReferenceTarget` instances will only have the `identifier` field set. Then, whenever invoking `someNode.someReference.target.node`, the node will be loaded from the node repository with all its reference by names not containing the corresponding node - but the only the identifier.

Also, with the current setup, I think replacing `PossiblyNamed` with `Node` would simplify the development and still meet our current use of reference by names. In order to handle unnamed references - if required - we could have another kind of reference without name.

```
class ReferenceTarget<NodeTy: Node>(val identifier: String, val node: NodeTy? = null)
class Reference<NodeTy: Node>(val target: ReferenceTarget<NodeTy>)
class ReferenceByName<NodeTy: Node>(val name: String, target: ReferenceTarget<NodeTy>): Reference<NodeTy>(target)
```

Do you think this makes sense?

ftomassetti:
Todo:
- [ ] Update to latest changes in #301 
- [ ] Verify if this PR becomes obsolete

ftomassetti:
Lorenzo and I discussed it: in the ReferenceByName we should not use the Symbol Description but just the identifier (for the cases in which we do not have the node already)

ftomassetti:
With the changes in this PR I got the [first functional test involving symbol resolution](https://github.com/Strumenta/starlasu-lionweb-repository-client/blob/48c36aad9f6b0b5ba394e00f0f97119258e1db18/starlasu-client/src/functionalTest/kotlin/TodoFunctionalTest.kt#L120) to work.

This PR depends on a PR in LionWeb Java (https://github.com/LionWeb-io/lionweb-java/pull/135)

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
```diff
@@ -1,13 +1,13 @@
-package com.strumenta.kolasu.lionweb
+package com.strumenta.kolasu.ids
```
ftomassetti, line 1: Changes in this file should disappear when rebasing is completed

core/src/main/kotlin/com/strumenta/kolasu/ids/SourceIdProvider.kt
```diff
@@ -1,7 +1,8 @@
-package com.strumenta.kolasu.lionweb
+package com.strumenta.kolasu.ids
```
ftomassetti, line 1: Changes in this file should disappear when rebasing is completed

core/src/main/kotlin/com/strumenta/kolasu/ids/StructuralNodeIdProvider.kt
```diff
@@ -0,0 +1,32 @@
+package com.strumenta.kolasu.ids
```
ftomassetti, line 1: Changes in this file should disappear when rebasing is completed

gradle.properties
```diff
@@ -9,6 +9,6 @@ org.gradle.jvmargs=-XX:MaxMetaspaceSize=512m
 jvm_version=1.8
 junit_version=4.13.2
 gson_version=2.10.1
-lionwebVersion=0.2.6
+lionwebVersion=0.2.9
```
ftomassetti, line 12: Changes in this file should disappear when rebasing is completed

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StructuralLionWebNodeIdProvider.kt
```diff
@@ -1,36 +1,22 @@
 package com.strumenta.kolasu.lionweb
 
+import com.strumenta.kolasu.ids.ConstantSourceIdProvider
```
ftomassetti, line 3: Changes in this file should disappear when rebasing is completed

lionweb/src/test/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverterTest.kt
```diff
@@ -17,6 +17,10 @@ class LionWebModelConverterTest {
     {
       "version": "1",
       "key": "com-strumenta-SimpleLang"
+    },
```
ftomassetti, line 20: Changes in this file should disappear when rebasing is completed

ftomassetti:
Hi @loradd , with this PR I verified that I can make the functional test on symbol resolution in https://github.com/Strumenta/starlasu-lionweb-repository-client pass

Perhaps the most relevant point for you to review is the introduction of a nodeIdProvider fields in `scope`, `ScopeDescription`, and `DeclarativeScopeProvider`

Other points are about the nullability of references

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -35,7 +38,9 @@ fun scope(
  *
  **/
 class ScopeDescription(
-    private val ignoreCase: Boolean = false
+    private val ignoreCase: Boolean = false,
+    private val nodeIdProvider: NodeIdProvider = StructuralNodeIdProvider()
```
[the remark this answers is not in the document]
ftomassetti: I think you are right. Removing it, the functional test still work so it was not necessary

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -88,7 +93,7 @@ class ScopeDescription(
         ignoreCase: Boolean,
         init: ScopeDescriptionApi.() -> Unit
     ) {
-        this.parent = scope(ignoreCase, init)
+        this.parent = scope(ignoreCase, nodeIdProvider, init)
```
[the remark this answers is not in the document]
ftomassetti: Removed also here

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/provider/declarative/DeclarativeScopeProvider.kt
```diff
@@ -46,16 +48,19 @@ inline fun <
  * ```
  **/
 open class DeclarativeScopeProvider(
+    open var nodeIdProvider: NodeIdProvider = StructuralNodeIdProvider(),
```
loradd, line 51: I think this is now unused and could be removed

loradd:
Awesome! I think we can merge this PR once done with the comments 👍

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/provider/declarative/DeclarativeScopeProvider.kt
ftomassetti: You are right: removing it (and remove it from the DeclarativeScopeProviderRule, as well, because otherwise code did not compile) everything still seems to work

ftomassetti:
Thank you @loradd , I applied the changes requested and as tests pass I will merge

PR #311, opened 2024-03-12
feat: introduces Symbol Provider DSL using @DslMarker annotations to limit scopes within lambdas

loradd:
feat: introduces Symbol Provider DSL using @DslMarker annotations to limit scopes within lambdas


loradd:
In this PR, we also modify the Symbol Description API to support arbitrary values as Symbol Description properties. That is, the API now supports including properties as follows:
```kotlin
symbolProvider {
    symbolFor(SomeNode::class) {
        include("someProperty", UUID.randomUUID().toString()) // for example
        include("anotherProperty", "anything that can be converted to a value description")
    }
}

```

semantics/src/test/kotlin/com/strumenta/kolasu/semantics/SymbolResolutionWithSRITest.kt
```diff
@@ -104,20 +97,10 @@ class SymbolResolutionWithSRITest {
...
-                    return SymbolDescription(
-                        node.name,
-                        id,
-                        emptyList(),
-                        mapOf("name" to StringValueDescription(node.name))
-                    )
-                } else {
-                    return null
-                }
+
```
ftomassetti, line 100: Nice!

ftomassetti:
I will wait for the PR to be marked as ready before reviewing it

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -1,67 +1,145 @@
...
+import com.strumenta.kolasu.validation.Issue
+import com.strumenta.kolasu.validation.IssueSeverity
+import com.strumenta.kolasu.validation.IssueSeverity.ERROR
+import com.strumenta.kolasu.validation.IssueSeverity.INFO
+import com.strumenta.kolasu.validation.IssueSeverity.WARNING
 import kotlin.reflect.KClass
-import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.isSuperclassOf
 
+// TODO merge into NodeIdProvider.kt file
```
ftomassetti, line 12: Should this TODO be solved before merging?

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -1,67 +1,145 @@
...
-open class DeclarativeNodeIdProvider(
-    vararg rules: DeclarativeNodeIdProviderRule<out Node>
-) : NodeIdProvider {
-    private val rules: List<DeclarativeNodeIdProviderRule<out Node>> = rules.sorted()
-
-    override fun id(kNode: Node): String {
-        return this.rules.firstOrNull { it.canBeInvokedWith(kNode::class) }?.invoke(this, kNode)
-            ?: throw RuntimeException("Cannot find rule for node type: ${kNode::class.qualifiedName}")
-    }
+fun nodeIdProvider(
```
ftomassetti, line 19: Regarind NodeIdProviders, I have been doing some reflections today that I put in this document: https://docs.google.com/document/d/1beGyoVn6t8lrivs77_0P002xEXUGvXd3tCm1uZJQZJA/edit#heading=h.brcynsfenjhr

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -1,67 +1,145 @@
...
+ * rule for a specific node type.
  **/
-class DeclarativeNodeIdProviderRule<NodeTy : Node>(
+@NodeIdProviderDsl
+sealed interface NodeIdProviderConfigurationRuleApi {
+    /**
+     * Add an information [message] during the rule execution.
+     * @param message the information message content
+     **/
+    fun info(message: String)
```
ftomassetti, line 56: Would it make sense to group these messages in an interface called Logging, or something similar, and then have this interface extend that one?

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
```diff
@@ -1,67 +1,145 @@
...
+    private fun issue(message: String, severity: IssueSeverity) {
+        this.issues.add(Issue.semantic(message, severity, this.context.node.position))
     }
 
-    override fun compareTo(other: DeclarativeNodeIdProviderRule<*>): Int {
+    override fun compareTo(other: ConfigurableNodeIdProviderRule<*>): Int {
         return when {
             this.nodeType.isSuperclassOf(other.nodeType) -> 1
-            this.nodeType.isSubclassOf(other.nodeType) -> -1
+            other.nodeType.isSuperclassOf(this.nodeType) -> -1
```
ftomassetti, line 118: Can we use also interfaces as nodeTypes? If so, would this work?

core/src/main/kotlin/com/strumenta/kolasu/semantics/ScopeProvider.kt
```diff
@@ -91,6 +91,7 @@ class ScopeProviderConfiguration(
 
 // builder
 
+@Deprecated("The corresponding component in the semantics module should be used instead.")
 fun scopeProvider(init: ScopeProviderConfiguration.() -> Unit) = ScopeProviderConfiguration().apply(init)
```
ftomassetti, line 95: Can't we just remove it?

gradle.properties
```diff
@@ -1,4 +1,5 @@
-version=1.5.49-SNAPSHOT
+#version=1.5.49-SNAPSHOT
+version=1.5.49-lsp-experiment-35
```
ftomassetti, line 2: We should get back to version=1.5.49-SNAPSHOT

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
```diff
@@ -4,133 +4,148 @@ import com.strumenta.kolasu.model.Node
...
-            }
-            identifier != null -> {
-                reference.identifier = identifier
-            }
-            else -> {
-                this.parent?.resolve(reference)
-            }
-        }
-    }
+    private val namesToSymbolKeys: MutableMap<String, String> = mutableMapOf()
```
ftomassetti, line 28: Can't I have more than one symbol with the same name in the scope? For example, overloaded methods

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -3,46 +3,173 @@ package com.strumenta.kolasu.semantics.symbol.description
...
+ * The description of a [Node] from the outside,
+ * i.e. from other trees not containing the node.
+ * @property name the name of the symbol
+ * @property identifier the identifier of the described node
+ * @property type a description of the type of the described node
+ * @property fields the visible fields of the described node
+ *
+ * @author Lorenzo Addazi <lorenzo.addazi@strumenta.com>
+ **/
 data class SymbolDescription(
```
ftomassetti, line 20: I would not make SymbolDescription as a Node, because it is not part of an AST and Kolasu nodes represent only parts of the AST. Is it possible to do this change or would that have too much impact?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -3,46 +3,173 @@ package com.strumenta.kolasu.semantics.symbol.description
...
 ) : Node(), Named
 
+/**
+ * The type description of a symbol.
+ * @property name the name of the type
+ * @property superTypes possible super types
+ *
+ * @author Lorenzo Addazi <lorenzo.addazi@strumenta.com>
+ **/
+data class TypeDescription(
```
ftomassetti, line 34: I would also make TypeDescription not nodes.
Also, for which purpose do we need TypeDescriptions?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
```diff
@@ -3,46 +3,173 @@ package com.strumenta.kolasu.semantics.symbol.description
...
+            type?.allSuperclasses?.any { this.isSuperTypeOf(it) } ?: false
+    }
+
+    /**
+     * Checks whether the described type is a supertype of the given [type].
+     *
+     * @param type the possible subtype
+     * @return true if [type] is a subtype, false otherwise
+     **/
+    fun isSuperTypeOf(type: TypeDescription?): Boolean {
```
ftomassetti, line 78: Can't we have rules for super typing and sub typing that are language specific? For example generics, variance, contra-variance, etc.?

semantics/src/test/kotlin/com/strumenta/kolasu/semantics/scope/ScopeProviderTest.kt
```diff
@@ -0,0 +1,49 @@
+package com.strumenta.kolasu.semantics.scope
+
+import org.junit.Test
+
+class ScopeProviderTest {
+
+    @Test
+    fun testScopeFor() {
+//        val internalNode = InternalNode(
```
ftomassetti, line 9: We should revise this one

ftomassetti (commented):
I think we miss tests for type calculation

ftomassetti:
Could you merge or rebase this on master?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: Type descriptions are used to describe the type of the node described by a symbol - its name and possible supertypes. These are used when reasoning over the type of symbol without loading the corresponding node. I think we could define these without extending Node, is there any issue related to these being Node instances?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: We discussed and decided to define SymbolDescription as nodes some weeks ago and the internal implementations have been modified with this idea in mind. This could be avoided but it would have impacts on the Symbol Provider configuration API.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
loradd: Yes, multiple symbols for the same symbol key are possible - both in the local and external symbols map. This map is only keeping track of the original name and the corresponding key. For example, if I have a class 'Person' and a variable 'person' within a case-insensitive scope description, there will be two entries in this map: 'Person' -> 'person' and 'person' -> 'person'. The actual symbols will both be associated with the key 'person'. This allows us to keep track of the original names, which are used in features like code completion. Otherwise, we would lose this information when converting names to keys, e.g. Person would become person and be suggested in the wrong way.

loradd: Considering the overloaded methods support: it would require a specific symbol describing methods. I think this is a language-specific feature. In the scope description, we can have multiple symbols associated to a given name. However, we can only have one symbol per type in this list. Hence, the multiple overloaded implementations of a method should be represented inside the function symbol. Am I wrong?

core/src/main/kotlin/com/strumenta/kolasu/semantics/ScopeProvider.kt
loradd: We decided to use annotations to maintain back compatibility. I completely agree with removing these classes, perhaps once all tests will be migrated.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: Given the definition `NodeTy: Node`, I do not think it would work with interfaces. In order for this to work with interfaces, we would need to remove the type constraint.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: I am fine with this, we could also just define a set of extension methods over `MutableList<Issue>` and use them with the rule argument `issues`.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: Could not we have a concept of identifier in Kolasu which is not necessarily depending on LionWeb? We are using it to keep track of the correspondence between symbols and nodes, for example. Global symbol resolution should also be usable in context where information is kept in-memory - e.g. through an InMemorySymbolRepository - rather than LionWeb repositories. Could not LionWeb be just one possible option to store the information?

ftomassetti: I think that, for now, LionWeb is the only place where we need to use a node identifier. We may have later on situation where we may need a different way to handle identifiers but:
1. I do not know about such use cases
2. It may be the case that we can use the same logic to attribute identifiers that happen to work for LionWeb

So I would lean toward not generalizing at this stage

For the case where we use symbol repository in memory, wouldn't the same ID generated for LionWeb work too?

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: Ok, both works for me

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: Right, let's forget about this then. This may change in Kolasu 1.6 where we have an interface to represent anything that can be a Node

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
ftomassetti: > Hence, the multiple overloaded implementations of a method should be represented inside the function symbol. Am I wrong?

I am not sure I understand this part

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: Sorry, I do not remember why we decided to make SymbolDescriptions to extend Node. Do you remember the reason? Is that reason still true?

In order not to delay merging this PR I would then write down the rationale for making SymbolDescription extend Node, and add a note that this could be later reconsidered

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: Oh, I see. Sorry, I thought this represented types calculated for expressions. I instead understand that they represent instead the type of the Node (e.g., the AST class). In Kolasu 1.6., we have classes representing the metamodel (e.g., Concept) and we could use that for this goal. 

> is there any issue related to these being Node instances?

The issue is that a Node should represent an AST Node. Using it for other things could lead to confusion

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: I will create a _IssueExtensions_ file in the _semantics_ module introducing such extension methods. As concerns those used in _DeclarativeNodeIdProvider_, if we keep it, the methods remain separated as those used in _SemanticProvider_ subclasses are creating _semantic_ issues by default. Does this make sense?

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
loradd: I agree, this is just a temporary solution that will radically change in Kolasu 1.6

core/src/main/kotlin/com/strumenta/kolasu/semantics/ScopeProvider.kt
loradd: I will migrate all tests and remove the old Symbol Resolution API classes.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: Initially, we thought it would have made integration with LionWeb easier - re-use logic for storing/loading Kolasu nodes. I don't know if that is still the case, but it would still be valid in case we would decide to store SymbolDescriptions inside a LionWeb repository. For the same reason, we defined the _Symbol Description Language_. Defining symbol descriptions as nodes is also useful as all traversal and other Kolasu APIs can be reused. Is there any document where we decided that Kolasu nodes only represent AST nodes?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
loradd: I think keeping these as Nodes would allow us to re-use most of the Kolasu API for navigating, tracking positions, representing references. Is there a document where we decided that Kolasu nodes can only represent nodes strictly contained in the AST? I feel this is not really the current practice, e.g. nodes have also been used to represent types or external symbol representations in past discussions/projects.

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: > Ok, should we remove the DeclarativeNodeIdProvider and re-introduce it in the future in case we need it?

Yes

> The current symbol provision APIs work with a generic NodeIdProvider interface, so as long as the LionWeb StructuralNodeIdProvider works for in-memory cases everything should be fine.

Perfect, thank you

core/src/main/kotlin/com/strumenta/kolasu/ids/DeclarativeNodeIdProvider.kt
ftomassetti: > I will create a IssueExtensions file in the semantics module introducing such extension methods. 

Perfect

> As concerns those used in DeclarativeNodeIdProvider, if we keep it, the methods remain separated as those used in SemanticProvider subclasses are creating semantic issues by default. Does this make sense?

Makes sense

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
ftomassetti: From my point of view symbol resolution should figure out, among other things, which method get called. And that means also distinguishing between overloaded variants of methods with the same name. So after symbol resolution is performed I would expect the reference in a method call to point to the right method declaration. In this way other systems can just consume that information without having to know anything on the method resolution rules. For example, some editor could just use that information to implement a "go to declaration" functionality. Symbol resolution should fine the information necessary to perform the resolution by looking at the symbol definition.

> Otherwise, how should handle cases where multiple symbols of the same type and name are contained in a given scope? 

I was thinking to have a symbol for each overloaded variant, describing the signature of that method and symbol resolution would consider the information to pick the right symbol. If the actual symbol resolution is delegated to FunctionSymbol, then the user needs to know about it and calling the appropriate method. 

> Otherwise, how should handle cases where multiple symbols of the same type and name are contained in a given scope? 

Considering not only the name but the whole symbol description. Otherwise why do we need to have any information but the name in the Symbol Description? The information is there only to be consumed during symbol resolution, or at least this was my understanding

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: > Initially, we thought it would have made integration with LionWeb easier - re-use logic for storing/loading Kolasu nodes. I don't know if that is still the case,

I think this is not anymore the case. Sorry for the confusion. We can implement serialization to LionWeb (and deserialization) rather easily, without the need to force extending Node for that goal

> Defining symbol descriptions as nodes is also useful as all traversal and other Kolasu APIs can be reused

I see the point

> Is there any document where we decided that Kolasu nodes only represent AST nodes?

No, and there should be one.

I think that on a practical level is sometimes tempting to make something a Kolasu Node, to reuse the APIs we have. But on a conceptual level this causes confusion as we end up having Node indicating _mostly_ AST nodes and sometimes other things, for which the class Node have features that makes sense to use for proper AST nodes, but not for other Nodes.

Maybe in Kolasu 1.6 we could have two interfaces: one called GenericNode and one called ASTNode extending GenericNode, but I do not see clearly what abstraction would GenericNode represent

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/symbol/description/SymbolDescription.kt
ftomassetti: I added a comment about this above. I agree this should be something well documented and defined, while now it is a sort of grey area

ftomassetti: P.S. I understand on this topic things are changing (based on what came up working on the Code Insight Studio and applying some ideas in practice) and that we today had limited time to discuss all aspects of it during the technical meeting. I am sorry for any complication caused by these updates and I am happy to have a meeting to discuss these things with more calm, if it is useful

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scope/description/ScopeDescription.kt
loradd: At the moment, the symbol resolver traverses the AST and aggregates all references. For each reference, it computes the corresponding scope and tries to resolve the reference using this. The resolution is generic and uses:

- The name of the reference by name instance;
- The type of the reference by name definition;

It absolutely makes sense to consider the signature in cases of overloaded functions, for example. I am not sure, however, how this could be implemented in a generic way. It looks to me like we might need some mechanism to specify the matching criteria. For example, in the case of a variable, it might be the name. In case of functions, it could integrate parameter type information as well.

The actual sequence of symbol resolution steps might change from node to node. For example, when resolving symbols for a function call expression, we might want to do it on its argument expressions first. I feel we could introduce support to specify resolution steps, for example:

```kotlin
resolve(InvocationExpr::class) { (invocationExpr) ->
   // resolve symbols in arguments first
   invocationExpr.arguments.forEach { argument -> resolve(argument) }
   // compute the scope for the invocationExpr::function property
   val scope = scopeProvider.scopeFor(invocationExpr.function)
   // use the scope to obtain a list of candidates
   // select the most appropriate candidate according to some criteria
   invocationExpr.function.identifier = /* the identifier of the selected candidate */
   // we could also differentiate between externalCandidates and localCandidates
   invocationExpr.function.referred = /* the node result from asking local candidates */
}
```

The workflow would consist in:
- Defining a Symbol provider to specify which nodes can be referenced from the outside;
- Defining a Scope provider to specify what is visible from a given Node;
- Defining a Type provider to specify the type for a given Node (considers references as resolved);
- Defining a Symbol resolver to specify how to resolve symbols;

In the definition of a symbol resolution rule, we might use a type provider to retrieve the type of a given node - with the assumption that all required symbols have been resolved if possible. A scope provider to return back the scope description from a given node. The scope provider can use the type provider, but the type provider cannot use the scope provider. Symbol resolution rules coordinate and make sure that symbols are resolved before being used. Does this make sense?

loradd:
Replaced by #347

Issue #291, opened 2024-01-15
KolasuParser, assignParents and AST transformers


alessiostalla:
This issue in Tylasu https://github.com/Strumenta/tylasu/issues/57 made me realise that we're calling `assignParents` in KolasuParser and similar classes, while it's unnecessary most of the time since we're using AST Transformers that set the parent already. If KolasuParser was aware of AST transformers, we could include a method e.g. `setupASTTransformer` that could return either:
- a transformer: the parser would store it, and not call `assignParents`;
- null: the parser would infer that no transformer is being used, and keep calling `assignParents` as it does now.

alessiostalla:
BTW this would also ensure that all parsers correctly pass the Source down to the AST transformer. Currently, one has to remember to do it manually, and we tend to forget it.

ftomassetti:
In Kolasu 1.6 we will be able to automatically set the parent, when the kotlin compiler plugin will be used, so `assignParents` will not be necessary anymore

JiaLiangC:
Are there any plans to maintain a Java version of the "antlr ParseTree to AST" tool?

ftomassetti:
> Are there any plans to maintain a Java version of the "antlr ParseTree to AST" tool?

Hi @JiaLiangC can you provide more details about what you have in mind?

PR #323, opened 2024-04-16
Update on Node IDs calculation

ftomassetti:
Based on the reflection in the [Node IDs in the Code Insight Studio](https://docs.google.com/document/d/1beGyoVn6t8lrivs77_0P002xEXUGvXd3tCm1uZJQZJA/edit?usp=sharing) document we update the logic for calculating IDs.

The typical approach is:
* to use Semantic Node IDs, when available
* to use Positional Node IDs, in the other cases


core/src/main/kotlin/com/strumenta/kolasu/ids/Coordinates.kt
loradd, line 1: What are Coordinates? What is a containment in Kolasu? Are we sure we are not mixing up LionWeb concepts into Kolasu?

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
```diff
@@ -10,34 +9,44 @@ import com.strumenta.kolasu.model.Node as KNode
...
 
-sealed class Coordinates
-
-object RootCoordinates : Coordinates()
-
-data class NonRootCoordinates(
-    val containerID: String,
-    val containmentName: String
-) : Coordinates()
+interface SemanticNodeIDProvider {
```
loradd, line 49: What is a SemanticNodeId? What semantics are we referring to here?

loradd (commented):
I am not sure I catch all the details but I have the feeling that node identifiers management mechanisms are getting a bit too complex and tailored for LionWeb. I suggest we add another reviewer to ensure a good feedback.

core/src/main/kotlin/com/strumenta/kolasu/ids/Coordinates.kt
ftomassetti: "Coordinates" indicate where a Node is stored. A root node will have RootCoordinates, a node that is not a root node will have NonRootCoordinates. NonRootCoordinates indicates the parent (or its id), the containment property and index.

For example, if I have this:
```
class ClassDeclaration(val name: String, val members: List<Member>) : Node()
abstract class Member: Node()
class FieldDeclaration(val name: String): Node()

val f1 = FieldDeclaration("F1")
val f2 = FieldDeclaration("F2")
val f3 = FieldDeclaration("F3")
val myClass = ClassDeclaration("foo", listOf(f1, f2, f3))
```

* myClass will have RootCoordinates
* F1 will have NonRootCoordinates(parent=myClass, property="members", index=0)
* F2 will have NonRootCoordinates(parent=myClass, property="members", index=1)
* F3 will have NonRootCoordinates(parent=myClass, property="members", index=2)

I will add comments to the Coordinates classes.

The idea of containment is present both in LionWeb and Kolasu, but it is true that in Kolasu is less used and formalized. In Kolasu we had for a long time these methods to indicate the containment:

```
fun Node.containingProperty(): PropertyDescription?
fun Node.indexInContainingProperty(): Int?
```

In Kolasu 1.6 we will have better support for this

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
ftomassetti: The idea is that some nodes have an identity that is relevant for the semantic of the language, which typically means "one can refer to them". For example, Class Declarations and Method Declarations have this "semantic identity". A package declaration or an expression have not.

The semantics API permit to attribute an ID to nodes that are relevant for symbol resolution, i.e., for nodes with a semantic identity. From the point of view of the symbol resolution other nodes do not need an id.

From the point of view of LionWeb every Node needs an ID, but for nodes which do not have a semantic identity the ID is based on where they are in the AST.

So we distinguish between nodes with a semantic ID (in practice, node that get an ID from the semantic API you created) and not which get a positional ID (so one calculated based on their position on the AST, so basically their coordinates)

core/src/main/kotlin/com/strumenta/kolasu/ids/CachingNodeIDProvider.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.strumenta.kolasu.model.Node
+import java.util.IdentityHashMap
+
+/**
+ * The Node ID of most nodes is a Positional Node ID. Positional Node IDs are calculated by considering the Node ID of
+ * the parent,so that the Node IDs of all ancestors are calculated again and again.
+ * By using a cache we can avoid that.
+ */
+class CachingNodeIDProvider(val wrapped: NodeIdProvider) : BaseNodeIdProvider() {
```
alessiostalla, line 11: I've learnt that whenever I see a cache, without a method to clear it, my spider sense goes off..

core/src/main/kotlin/com/strumenta/kolasu/ids/Coordinates.kt
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Coordinates indicating the root of the AST.
+ */
+object RootCoordinates : Coordinates()
+
+/**
+ * Coordinates indicating the position of all nodes but the root of the AST.
+ */
+data class NonRootCoordinates(
```
alessiostalla, line 35: Would it make sense to replace "NonRoot" with "InternalNode", "Child", or other similar terminology?

core/src/main/kotlin/com/strumenta/kolasu/ids/SourceIdProvider.kt
```diff
@@ -24,15 +25,15 @@ abstract class AbstractSourceIdProvider : SourceIdProvider {
 
 class SimpleSourceIdProvider(var acceptNullSource: Boolean = false) : AbstractSourceIdProvider() {
```
alessiostalla, line 26: What's the purpose of this class? In the case of the EGL parser, I had problems (now solved) because it wasn't setting the source of the AST. Also see https://github.com/Strumenta/kolasu/issues/291

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
```diff
@@ -10,34 +9,44 @@ import com.strumenta.kolasu.model.Node as KNode
...
-            NonRootCoordinates(this.id(kNode.parent!!), kNode.containingProperty()!!.name)
+            positionalID(kNode)
         }
     }
-}
 
-interface IDLogic {
-    fun calculatedID(coordinates: Coordinates): String
+    private fun positionalID(kNode: Node): String {
+        return StructuralNodeIdProvider().apply { parentProvider = this }.caching().id(kNode)
```
alessiostalla, line 45: Isn't the point of caching lost here, since we're creating a new structural id provider for each request?

core/src/main/kotlin/com/strumenta/kolasu/ids/SourceIdProvider.kt
```diff
@@ -24,15 +25,15 @@ abstract class AbstractSourceIdProvider : SourceIdProvider {
...
     override fun sourceId(source: Source?): String {
-        if (source is IDLogic) {
-            return source!!.calculatedID(RootCoordinates)
+        if (source is SemanticNodeIDProvider) {
+            return source!!.semanticID(TODO())
```
alessiostalla, line 29: Currently I don't see any Source implementation which is also a SemanticNodeIDProvider, so what's the point of this branch?

alessiostalla (approved):
I don't know how this impacts semantics / symbol resolution so @loradd may want to comment about that. While generally I'm not enthusiastic about Lionweb requiring explicit IDs for each node, I think some form of id mechanism will be needed whatever storage solution we adopt; and if the complexity remains confined to the Lionweb integration, I'm OK with this solution.

core/src/main/kotlin/com/strumenta/kolasu/ids/CachingNodeIDProvider.kt
ftomassetti: Good point. This is intended to be short-lived but I will add a method to clear it.

The idea is that when traversing a tree:
A
|          \
B1       B2
|    \       |     \
C1 C2  C3 C4

We will end up calculating the node ID of A 7 times (once to set the node ID of A, then when setting the one of the other six nodes, that all depend on the node ID of A). So the idea was to use a CachingNodeIdProvider instance for processing an AST and then throw it away.

core/src/main/kotlin/com/strumenta/kolasu/ids/Coordinates.kt
ftomassetti: I will rename it into `InternalCoordinates`. Does this work?

core/src/main/kotlin/com/strumenta/kolasu/ids/SourceIdProvider.kt
ftomassetti: For nodes we calculate their IDs besides on two strategies:
- "Semantic Node IDs" for nodes that represent something with a proper identity, typically something you can refer to
- "Positional Node IDs" for all other nodes. Now when calculating the Positional Node ID of a root AST node we use an id derived from its source. This class is the one that given a source gives us back an ID we can use to produce the Node ID of the root of an AST having that source

Eventually I think that Sources should produce that ID on their own and we could remove this class

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
ftomassetti: Probably yes. I will remove it

core/src/main/kotlin/com/strumenta/kolasu/ids/SourceIdProvider.kt
ftomassetti: Yes, that wrong. I am removing it

core/src/main/kotlin/com/strumenta/kolasu/ids/NodeIdProvider.kt
loradd: Thanks for the explanation.

In the symbol resolution API, we can define a _SymbolProvider_ to specify which nodes should be visible from the outside and what should be visible - i.e. a _SymbolDescription_. The identifier is stored in _SymbolDescription_ instances to have the possibility to retrieve the full _Node_ instance from the repository - i.e. _LionWebRepository_ for example. 

In this context, I think we could only keep the positional identifier. Nodes that we do not want to make visible from the outside - e.g. expressions - will not have a corresponding _SymbolDescription_ and will not be stored in the _SymbolRepository_. What do you think?

When creating a symbol description, we can then use one of these IdProvider classes - perhaps an interface if there is any or directly the default identifier provider that we plan to use. The only feature that is expected by the symbol provider and other semantics components is to be able to generally obtain the same ID for a given node in a deterministic way. I think this is covered by your implementation. 

To summarise, I think we could avoid having an additional concept of semantic identifier. Am I missing something?

ftomassetti: > Nodes that we do not want to make visible from the outside - e.g. expressions - will not have a corresponding SymbolDescription and will not be stored in the SymbolRepository. What do you think?

Yes, absolutely.

But before you write "In this context, I think we could only keep the positional identifier", while it would be the nodes with semantic IDs and not positional IDs to end up in the Symbol Repository

> When creating a symbol description, we can then use one of these IdProvider classes - perhaps an interface if there is any or directly the default identifier provider that we plan to use. 

Actually I was thinking the other way around: you have a system to calculate the semantic IDs. The more general NodeIdProvider should first call your system and if your system can produce an ID that would be used. Otherwise, the positional ID would be generated. This is what the class `CommonNodeIdProvider` does, trying to call `DeclarativeNodeIdProvider` first. When `DeclarativeNodeIdProvider` does not return an ID, because we passed to it an expression, then the positional ID would be calculated.

ftomassetti:
Based on the discussion with @loradd I am merging this, and remain available to help updating [the pending PR](https://github.com/Strumenta/kolasu/pull/311)

PR #325, opened 2024-04-18
Handle deserialization of properties NOT set at construction time

### LINKED ISSUE #324: LionWebModelConverter only populates nodes using constructor parameters
Most nodes are data classes and have all properties as constructor parameters, but if you have a node where some children or attributes are not set at construction time, they are silently left out (= null). This has actually happened.

ftomassetti:
We mostly use data classes that set all properties at construction time. However, we should also initialize the properties that are not set at construction time.

Fix #324


ftomassetti:
Going ahead and merging this PR, so that we can include it in the next release

PR #347, opened 2024-06-03
Feature/semantic enrichment

loradd:
Replaces #311


mctaverna (approved):

mctaverna (approved):

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
```diff
@@ -0,0 +1,70 @@
...
+        this.values.remove(value)
+    }
+
+    override fun insert(value: Any) {
+        this.remove(value)
+        this.values.add(value)
+    }
+}
+
+// TODO consider limited buffer size
```
ftomassetti, line 70: Is this still relevant?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
```diff
@@ -0,0 +1,70 @@
+package com.strumenta.kolasu.semantics.indexing
+
+/**
+ * Abstract storage component for external symbols.
+ **/
+interface Index {
```
ftomassetti, line 6: Should this be called SymbolIndex?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Retrieve first value matching [predicate] from the index.
+     * @param predicate the predicate to match
+     * @return the associated value (exception if none)
+     **/
+    fun get(predicate: (Any) -> Boolean): Any {
```
ftomassetti, line 12: Should we use an interface representing Symbols instead of Any in this and successive methods?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexReader.kt
```diff
@@ -0,0 +1,75 @@
+package com.strumenta.kolasu.semantics.indexing
+
+import kotlin.reflect.KClass
+import kotlin.reflect.safeCast
+
+/**
+ * Component handling read operations over an index.
```
ftomassetti, line 7: It is not clear why we need this? Can't we just read directly from the index? Is this a sort of cache?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolverCache.kt
```diff
@@ -0,0 +1,41 @@
+package com.strumenta.kolasu.semantics.linking
+
+import com.strumenta.kolasu.model.ReferenceByName
+import kotlin.reflect.KProperty1
+
+/**
+ * Abstract storage for resolution results.
+ **/
+interface ReferenceResolverCache {
```
ftomassetti, line 9: Why do we need this? Can't we just change the references to store the new state?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexWriter.kt
```diff
@@ -0,0 +1,66 @@
...
+ * @property rules the index writer rules to use
+ **/
+class IndexWriter(
+    private val index: Index = InMemoryIndex(),
+    private val rules: Map<KClass<*>, (Any) -> Any?> = emptyMap()
+) {
+    /**
+     * Execute the index writer rule for the given [input].
+     * @param input the index writer rule input
+     * @return the output of the corresponding rule
```
ftomassetti, line 31: Does null indicate that no symbol has been produced for the given node?

ftomassetti (approved):
+1

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
loradd: Perhaps the documentation comment is confusing, the rationale behind the index interface is: to have a component responsible for storing external information/symbols/data, e.g. symbols, types, built-in things not stored in any AST, etc. Anything that should be visible from an AST but is not included in that AST. 

Given the rationale, I would not refer to Symbols only in this case. Using the index to store external symbols is one possible usage of this data structure. 

If this makes development easier somewhere else, we could perhaps define a SymbolIndex interface extending this one. But I am not sure about the advantage of this, any specific use case in mind?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
loradd: No, sorry, it slipped away :)

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexReader.kt
loradd: The index reader provides a read-only view over an index. This component is available when defining a reference resolver, type computer, scope computer, etc. The objective is to prevent writes and allows type-safe queries. 

The idea is that we can have an specific index - the actual storage depending on the implementation - and generic readers and writers providing advanced r/w features regardless of the underlying technology.

Does it make sense?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexWriter.kt
loradd: null is returns in two cases:

* there is no rule for the given input
* the rule produced a null value

A node is written into the index iff a rules exists and it produces a non-null value. The second case provides the possibility to have conditional rules - e.g. export classes only if public.

I will update the method documentation to make this clear.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolverCache.kt
loradd: The reference resolver returns back the result of resolving a reference, which is then used by the SymbolResolver to update references. The reference resolver can be used just to retrieve the target of a reference, regardless of it being local/global resolved/not resolved. For example, I could use it in a type computer to retrieve the target of a reference I want to compute the type for.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
alessiostalla: Maybe then it's not really an _Index_ but an _Environment_ or similar concept?
_Index_ to me suggests something that's optimized for certain search queries, e.g. partitioning the values into some sort of ordered tree.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scoping/ScopeComputerCache.kt
```diff
@@ -0,0 +1,37 @@
+package com.strumenta.kolasu.semantics.scoping
+
+/**
+ * Abstract storage for scope computation results.
+ * @author Lorenzo Addazi <lorenzo.addazi@strumenta.com>
+ **/
+interface ScopeComputerCache {
```
alessiostalla, line 7: Same, what is the "scope" of the cache, how long does it live?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolver.kt
```diff
@@ -0,0 +1,138 @@
...
+import com.strumenta.kolasu.semantics.indexing.IndexReader
+import com.strumenta.kolasu.utils.sortBySubclassesFirst
+import kotlin.reflect.KClass
+import kotlin.reflect.KProperty1
+import kotlin.reflect.safeCast
+
+/**
+ * Type-safe builder to create resolver instances.
+ * @param index optional index for external symbols
+ * @param cache optional storage to cache computations
```
alessiostalla, line 16: It's not really optional though

alessiostalla (approved):

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
loradd: > Maybe then it's not really an _Index_ but an _Environment_ or similar concept? _Index_ to me suggests something that's optimized for certain search queries, e.g. partitioning the values into some sort of ordered tree.

It is meant to be an interface where to store views of data, which details depend on the specific objective for which we are storing it. For example, we could have entries for node positions consisting of an identifier and a position value, or else. 

Aren't the specifics of data partitioning and query execution something that would belong to a specific implementation of the Index?

I would be fine with using another name like _Environment_, but then it might become unclear that we can also store partial views of things using custom index writer configurations. The point is that whether it is an environment, a cache or else depends on the implementation you choose or define and the writer configuration.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolverCache.kt
loradd: I am not sure I understand, what would be a use case for emptying this cache? This is used to avoid repeating computations during reference resolution, much like memoization was used before. Having the possibility to provide different implementations is only a plus for cases where we might want different solutions than in-memory maps.

In a similar way, a custom implementation of a resolution cache could handle allocation/deallocation logic accordingly. No policy is currently defined.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolver.kt
loradd: You're right, optional is probably not the best word. I will change to specify that if no index is provided, an initially empty in-memory one will be used. What do you think?

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scoping/ScopeComputerCache.kt
loradd: I will improve documentation for all caches. The idea is that how long it lives really depends on the implementation and specific use case. The in-memory map implementation is volatile, but persistent caches could be implemented as well (for something like LSIF).

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolver.kt
loradd: Commented

loradd: Modified documentation with:
```kotlin
@param index index for external symbols (default: [InMemoryIndex])
@param cache storage to cache computations (default: [InMemoryReferenceResolverCache])
```

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/scoping/ScopeComputerCache.kt
loradd: The documentation of the _ScopeComputationCache_ is now:

```kotlin

/**
 * Abstract storage for storing scope computation results.
 *
 * Scope computation caches are meant to support the scope
 * computer not to repeat calculations. While a persistent
 * version of a scope computation cache can be implemented,
 * it is meant to be volatile by default.
 *
 * Persistent implementations should design dedicated
 * mechanisms for emptying, allocating and deallocating the cache.
 *
 **/

```

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
ftomassetti: To me seems an over-generalized abstraction. My suggestion is to create generic abstractions when you have 3 different concrete cases. Without concrete examples the interface seems too "abstract" to me. I would not know when to use it and when not to use it. To me it seems a subset of the `Collection` interface.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexReader.kt
ftomassetti: To me this seems similar to the `Collection` interface (the immutable variant). I understand the "offering a subset of Index" part, but as Alessio I would think of it as the base interface from which Index would be derived. Alternatively it could be a specialization of Index which just throw exceptions when the writing methods are called. Similarly to Index seems to me to bee over-generic. It also does not seem to have any method specific to indexes. If anything it seems to me a "ReadOnlyCollection". Perhaps you need it because you need to use the PublishedAPI annotation. If that is the case I would explain that in the documentation. Maybe one could present it as a sort of ReadOnlyCollection needed exclusively to support the DSL API and not intended to be reused for other goals

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/linking/ReferenceResolverCache.kt
ftomassetti: I understand that the goal is not to perform resolution more than once.

What I do not understand is under which circumstances you could not just use the ReferenceByName itself as the cache: when asked to resolve a reference, if resolved is already set you just get the result from there. In case the resolved field is set to null, then yes, you would end up recalculating the reference (to find out it is null) each time. Is it to avoid this case that you use the cache? If so:
- I would explicitly explain this in the documentation
- I will open an issue about reflecting on how to distinguish between the states "Resolution not yet calculated" and "Resolution calculated and resulting in no match" in Kolasu 1.6

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexWriter.kt
ftomassetti: > I will update the method documentation to make this clear.

Good, thank you

Regarding IndexWriter, I have the same comments I have for IndexReader

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexWriter.kt
```diff
@@ -0,0 +1,66 @@
...
+     **/
+    fun write(input: Any): Any? {
```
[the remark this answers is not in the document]
ftomassetti: To me this seems a symptom that this is overly generic. If it was a class created to work with Nodes, by looking at the signature I could figure that out. If instead the signatures are about using `Any` then it makes not obvious what this class can be used for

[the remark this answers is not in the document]
loradd: This component is not meant to work with nodes only, but also types, symbols and any other things that might be stored in the index. 

I think it would be enough to have some documentation about the various components to make users understand what can be used and for what.

This and other comments give me the feeling that the whole solution for indexing should be redesigned. Perhaps we could postpone this into a later PR with a different architecture.

Otherwise, I am not sure how to proceed in addressing these comments and would accept suggestions/contributions.

loradd:
This PR contains the latest development to support semantic enrichment in Kolasu. 
Feel free to merge it when you consider it ready, commit changes, etc.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexReader.kt
loradd: I am not sure I understand why this looks similar to the _Collection_ interface. 
The current code is a proposal, feel free to change it as you wish.

alessiostalla: Well, a Collection has additional requirements I think, at least in Java, so it makes sense that one doesn't have to implement all Collection methods just to have access to the index/storage. That said, I too have the feeling that this design is a little too abstract. Maybe we could do without index readers and writers, and have the index present a writable view and a read-only view? I.e. `Index` by default is writable, but if you want to protect it from writes, you call a `.readOnlyView()` method on it (or access a `.readOnlyView` getter) that returns an instance wrapping the index and throwing exceptions on all methods that would mutate it. Kind of what happens with `Collection.unmodifiableCollection()` in Java. That's just a suggestion.

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
alessiostalla: I understand this index is a portal to the external world. So in my mind I'd call it `ExternalStorage` or something like that. And it makes sense that in the library there is only a "trivial" implementation of it, because implementations using LSIF or the LW Model Repo etc. do not belong in Kolasu Core. This is I think what other libraries/frameworks call an "SPI" (e.g. Hibernate), i.e. an extension point that you can inject into the library. Kind of the opposite direction of an API, that is meant to be called by consumers of the library, while the SPI is meant to be called by the library with a well-defined protocol. 
It makes sense to me now, but I was initially confused by the name "Index" that maybe comes from LSIF.
That said, I too think that index reader and writer are maybe a bit overkill, but I don't think that should stop this PR from being merged; we can always simplify the design later, if the API proves to be confusing, verbose, etc.

alessiostalla (approved):

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/Index.kt
ftomassetti: After discussion we decided to rename the package persistence and this class `Repository`

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexWriter.kt
ftomassetti: After discussion we decided to call this `RepositoryWriter`

semantics/src/main/kotlin/com/strumenta/kolasu/semantics/indexing/IndexReader.kt
ftomassetti: After discussion we decided to call this `RepositoryReader`

alessiostalla:
Why is this closed? I thought we had approved it!

loradd:
> Why is this closed? I thought we had approved it!

The approach seems to be incompatible with our current practices in defining ASTs and the planned architecture of large-scale migrations using the Code Insight Studio, LionWeb and the Symbol Resolution Index (SRI). 

The previous attempt using generic _SymbolDescription_ instances seems to be enough for EGL. I would suggest using the previous and more usable APIs for projects where no external symbols are needed/handled. This is an example project for the previous APIs: [Entity Parser](https://github.com/Strumenta/kolasu-entity-parser)

Issue #352, opened 2024-06-13
Support TODO placeholder nodes


alessiostalla:
Especially when transpiling, it can happen that we want to flesh out an AST but leave out some nodes that we don't yet generate, e.g.:

```
class A {
   /* translation of RPG field a */
   /* translation of RPG field b */
}
```

Currently we have no general way of doing that. Kolasu is in a good position to provide some support.

**Ideas and discussion items**

* **Kolasu 1.5 or 1.6** since we want this on current language modules, it makes sense that we add this to Kolasu 1.5, or should we wait for 1.6 to start being adopted?
  * In Kolasu 1.6, we could use generated error nodes with some annotation
* **Integration with AST Transformers** currently AST transformers have a concept of "generic node" which is not very usable because it doesn't work well with the type system, we could replace that with the insertion of placeholder nodes
  * This would better go in Kolasu 1.6 so we don't introduce breaking changes to 1.5
  * It would require some sort of factory to map node types to placeholder node types
  * My gut feeling is that this is overkill and we can just supply explicit rules for placeholders when needed

alessiostalla:
I did several experiments. It turned out that very little is needed in Kolasu. I will open a couple of PRs, one with minimal support, and another one based on the first one with an additional feature I added that proved to be non necessary in the end (but could still be useful).

PR #353, opened 2024-06-13
#352 "TODO" placeholder nodes for transpilers

alessiostalla:
We introduce (and showcase in tests) a simple "todo" marker node for parts of the AST that haven't been covered yet.


ftomassetti:
I think this is a quite challenging conceptual problem to solve.
I think we want to be sure to explore what the solution would mean in practice.

Perhaps I can start with describing what I expect to happen when building a transpiler.

One start defining some initial transformations. For example, one specifies that an EGL program must be mapped to a Java Class Declaration. To do this he must define the transformation and specify which parts of the EGL program end up producing the parts of the Java Class Declaration. For example, the name of the EGL program will be the name of the Java Class Declaration. Then we can say that each statement of the EGL program will end up in the body of a method named "main" under the Java Class Declaration. Something like:

```
this.registerNodeFactory(EGLProgram::class) { eglProgram ->
    JClassDeclaration(eglProgram.name, 
        declarations=mutableListOf(JMethodDeclaration("main", 
            translateList(eglProgram.statements)
        )))
}
```

Now, if the transpilation is triggered it will fail as we have not specified the translation for each EGL statement into a Java statement. 

We would like to change that and to get a partial translation. Ideally, the developer would not need to do any extra work, like creating special node subclasses or define temporary transformations to be later removed. They would just trigger the transpiler and get some code representing a partial transformations. For example:

```
class aProgram {
    void main() {
        /* Translation of EGL Statement 1 (EglIfStmt) */
        /* Translation of EGL Statement 2 (EglExprStmt) */
        /* Translation of EGL Statement 3 (EglExprStmt) */
    }
}
```

Later on the developer decides to go back and add the transformation from EglIfStmt into JavaIfStmt:
```
this.registerNodeFactory(EGLIfStmt::class) { eglStmt ->
    JavaIfStmt(
        translateCasted(elgStmt.condition), 
        translateCasted(eglStmt.thenBody, 
        translateOptional(eglStmt.elseBody)))
}
```

After triggering the transpiler once more he will now get this:

```
class aProgram {
    void main() {
        if (/* Translation of EGL Expression 1 (EglTrueExpr) */) {
            /* Translation of EGL Statement 4 (EglExprStmt) */
        }
        /* Translation of EGL Statement 2 (EglExprStmt) */
        /* Translation of EGL Statement 3 (EglExprStmt) */
    }
}
```

And so the work will produce, incrementally. It will be possible to involve the Client from the very beginning and show progress without first having to write hundreds of transformations.

The challenge is that we need to define at all times structurally valid ASTs for two reasons:
- Because otherwise we cannot instantiate them. For example, we may have AST nodes that _requires_ certain children to be present (for example the `thenBody` of a `JavaIfStmt` _cannot be null_). So we cannot temporarily pass _null_ or something to that effect. We always need to provide "something" to instantiate that AST node and that something is expected to have a given type (`JavaStmt` in this example)
- We need to produce valid AST nodes to pass them to the Code Generator, as it is the only mean we have to get code to display from the AST node we obtained through the transformers

So, yes, we need these sort of "placeholders" to be placed where the future AST nodes will go (once we define the corresponding transformations). In this way the Code Generator knows where to display them. For example, it will now that it needs to add the string representing the placeholder for the statements of our main method between the braces opening and closing the body of the method.

The fact is that, similarly to error nodes or proxy nodes, because of static typing we would need one subclass for each possible kind of node. This is impractical because it requires defining new node types just for the sake of representing these partial transformations, and we may have to define a lot of them. And it may be outright not possible as data classes cannot be extended and we use data classes for most AST classes.

For example, if I am translating an RPG subroutine to a method I may need to translate the body, but also the parameters:

```
this.registerNodeFactory(RPGSubroutine::class) { sub ->
    JavaMethod(
        sub.name,
        translateList(sub.params),
        translateList(sub.body))
}
```

With the approach suggested in this PR I would need to either:
1 - Define a subclass of JParameter. That requires changing the original definition of JParameter from a data class to an open, non data class (otherwise it cannot be extended). Then I will need to define a transformation from RPGParameter to TodoJParameter
2 - To reduce the number of Todo* classes, for all classes but expressions and statements we would not define the Todo* class, but we would just write something like:

```
this.registerNodeFactory(RPGSubroutine::class) { sub ->
    JavaMethod(
        sub.name,
        // change this translateList(sub.params once we have the transformation of params
        emptyList(),
        translateList(sub.body))
}
```

This is a bit error prone, and it does not work when an empty list or null cannot be used.

So what does it leaves us with?

Well, I don't know.

I can think of three possible alternatives. None seems perfect or ideal to me.

# Solution 1 - This solution

We go for this solution, with the idea that we want to create these placeholders-like classes for a few AST classes (expression, statement, and a couple of others). This means that partial transpilation involving other type of AST classes will not be possible. In some cases we will use null or empty list, and for cases in which this is not possible we will have to write right away the transformation, otherwise the transpiler will explode.
 
For cases where they are possible we will need to refer explicitly to those todo classes in the transformation. For example, we may write a transformation like this one:

```
this.registerNodeFactory(EGLProgram::class) { eglProgram ->
    JClassDeclaration(eglProgram.name, 
        declarations=mutableListOf(JMethodDeclaration("main", 
            eglProgram.statements.map { TodoJStmt(it) }
        )))
}
```

Later on, we will go back and change it to:
 
```
this.registerNodeFactory(EGLProgram::class) { eglProgram ->
    JClassDeclaration(eglProgram.name, 
        declarations=mutableListOf(JMethodDeclaration("main", 
            translateList(eglProgram.statements)
        )))
}
```

Note also that if we do that, and if we have transformations only for some statements but not for others, we will not get _any_ transformed statement. 

Alternatively we could write right away:

```
this.registerNodeFactory(EGLProgram::class) { eglProgram ->
    JClassDeclaration(eglProgram.name, 
        declarations=mutableListOf(JMethodDeclaration("main", 
            translateList(eglProgram.statements)
        )))
}
```

But then we would also need to "intercept" every possible subclass of statements we encounter and do not manage, otherwise the transpiler will crash:

```
this.registerNodeFactory(EGLIfStmt::class) { TodoJStmt(it) }
this.registerNodeFactory(EGLExprStmt::class) { TodoJStmt(it) }
this.registerNodeFactory(EGLWhileDoStmt::class) { TodoJStmt(it) }
...
```

Later on we will replace those transformations with the proper ones.

Problems I see:
* It provides a solution only for certain nodes but not all

# Solution 2 - Kolasu 1.6

We decide we cannot provide a "comfortable" alternative within Kolasu 1.5 and move as quickly as we can to Kolasu 1.6. In Kolasu 1.6 we solve this with a compiler plugin

Problems I see:
* It forces us to hurry up the migration to Kolasu 1.6, which may be risky

# Solution 3 - Dummy nodes + flag

We could use this approach for this and for generic nodes.

Basically, in the AST transformer we would define a method to be invoked when no matching transformation is found, the default case.

Let's imagine that, somehow, we can let this method know not only which node we could not transform but also which node class is expected to be produced (this is a big _if_).

```
this.registerDefaultCase { untranslatedNode, expectedNodeClass ->
    val translatedNode : Node = instantiate(expectedNodeClass)
    translatedNode.translationOf = unstranslatedNode
    translatedNode
}
```

The `instantiate` will instantiate any class we pass, by looking for a matching constructor (a constructor without cycles, see https://github.com/Strumenta/rpg-to-java-transpiler-product/blob/2ef93f6054b37bdb94bbcd5427e0eeaeb24d3c13/src/main/kotlin/com/strumenta/rpgtojava/demo/byexample/NodeCloning.kt#L31). Once a constructor is found it would produce a default value for each parameter:
- for lists -> emptyList
- for nullable values -> null
- for ReferenceByName -> ReferenceByName("", null)
- for node -> call instantiate again

The result would be a dummy node, with bogus value but the right type.

We would then change the `Node` class to have this field:

```
var translationOf: Node? = null
```

In the code generator we would check this field. When it is not null we will print:
`/* Translation of ${node.translationOf} */

We may be able to figure out the expected type through type inference, as we do with `translateCasted`,
whis is defined like:

```
fun <T> ASTTransformer.translateCasted(original: Any): T
```

But where we never pass `T`. Now, if we can make it inline the parameter can be reified wnad we could access `T` at runtime.

Problems I see:
* Not 100% sure it can be implemented
* It is a little hacky

alessiostalla:
Some thoughts.

1. I think _GenericNode_s were not a good design decision and we should remove them in Kolasu 1.6.
2. Before this PR I tried adding an "exception handler" to the AST transformers that could do this kind of node substitutions, but it proved to be complicated and didn't solve the problem elegantly anyway (becase you still needed concrete subclasses for each TODO variant). I had planned to provide this exception handler as a second PR on top of this one, but I don't think it's a good idea to add that extra complexity.
3. I like your third idea of instantiating the desired node anyway but somehow flagging it as "dummy". I wouldn't add a _translationOf_ field, as we have the _origin_ for a similar purpose already; instead I would either do as ANTLR does (adding an _Exception_ field) or using a special marker _Origin_ that represents a failed translation. If we do this, we won't need a special Todo marker interface anyway: the printer could just detect that condition and print a comment appropriate for the language that we're generating. It's true that we don't track the desired target type currently, but I don't think it would be hard to add.

Based on the above, I propose to:
1. Flag _GenericNode_ as deprecated (to be removed in Kolasu 1.6)
2. Drop the "exception handler" PR that I would have opened on top of this one
3. Close this PR and adopt idea number 3.

ftomassetti:
Some thoughts.

> I think _GenericNode_s were not a good design decision and we should remove them in Kolasu 1.6.

I think it is an interesting idea, but hard to concile with static typing. Could you open an issue about removing it and associate it with the milestone for Kolasu 1.6?

> Before this PR I tried adding an "exception handler" to the AST transformers that could do this kind of node substitutions,

Clever! I did not think about that

>I like your third idea of instantiating the desired node anyway but somehow flagging it as "dummy". I wouldn't add a translationOf field, as we have the origin for a similar purpose already; instead I would either do as ANTLR does (adding an Exception field) or using a special marker Origin that represents a failed translation.

Good idea! It is also much less invasive

> If we do this, we won't need a special Todo marker interface anyway: the printer could just detect that condition and print a comment appropriate for the language that we're generating. It's true that we don't track the desired target type currently, but I don't think it would be hard to add.

Perhaps, if we can retrieve the node type, it could be stored as a field of this special Origin subclass?

> Based on the above, I propose to:

> Flag GenericNode as deprecated (to be removed in Kolasu 1.6)
> Drop the "exception handler" PR that I would have opened on top of this one
> Close this PR and adopt idea number 3.

It sounds like a plan! Thank you Alessio, and I am always happy to discuss this feature if at any point you consider it useful to do so

PR #357, opened 2024-06-25
Support placeholder nodes in Lionweb too

alessiostalla:
We represent placeholder nodes in Lionweb as regular nodes with:

- A _PlaceholderNode_ annotation on them
- Their origin set to the _PlaceholderNode_'s origin (could also be null).
 
Note that this only works for dynamic nodes (see https://github.com/LionWeb-io/lionweb-java/issues/153)

### DIFF
--- a/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
+++ b/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
@@ -181,7 +181,7 @@ class LionWebModelConverter(
                                     } else {
                                         throw Exception(
                                             "MissingASTTransformation origin not supported on nodes " +
-                                                    "that are not AbstractClassifierInstances: $lwNode"
+                                                "that are not AbstractClassifierInstances: $lwNode"
                                         )
                                     }
                                 }
--- a/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
+++ b/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
@@ -168,7 +168,7 @@ class LionWebModelConverter(
                                     val targetID = myIDManager.nodeId(origin)
                                     setOriginalNode(lwNode, targetID)
                                 } else if (origin is MissingASTTransformation) {
-                                    if (lwNode is DynamicNode) {
+                                    if (lwNode is AbstractClassifierInstance<*>) {
                                         val instance = DynamicAnnotationInstance(
                                             StarLasuLWLanguage.PlaceholderNode.id,
                                             StarLasuLWLanguage.PlaceholderNode
@@ -180,7 +180,8 @@ class LionWebModelConverter(
                                         lwNode.addAnnotation(instance)
                                     } else {
                                         throw Exception(
-                                            "MissingASTTransformation origin not supported on non-dynamic node $lwNode"
+                                            "MissingASTTransformation origin not supported on nodes " +
+                                                    "that are not AbstractClassifierInstances: $lwNode"
                                         )
                                     }
                                 }
--- a/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
+++ b/lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
@@ -3,40 +3,20 @@ package com.strumenta.kolasu.lionweb
 import com.strumenta.kolasu.ids.IDGenerationException
 import com.strumenta.kolasu.ids.NodeIdProvider
 import com.strumenta.kolasu.language.KolasuLanguage
-import com.strumenta.kolasu.model.CompositeDestination
-import com.strumenta.kolasu.model.Multiplicity
-import com.strumenta.kolasu.model.Point
-import com.strumenta.kolasu.model.Position
-import com.strumenta.kolasu.model.PossiblyNamed
-import com.strumenta.kolasu.model.ReferenceByName
-import com.strumenta.kolasu.model.allFeatures
-import com.strumenta.kolasu.model.asContainment
-import com.strumenta.kolasu.model.assignParents
-import com.strumenta.kolasu.model.isAttribute
-import com.strumenta.kolasu.model.isContainment
-import com.strumenta.kolasu.model.isReference
-import com.strumenta.kolasu.model.nodeOriginalProperties
+import com.strumenta.kolasu.model.*
+import com.strumenta.kolasu.transformation.MissingASTTransformation
 import com.strumenta.kolasu.traversing.walk
-import io.lionweb.lioncore.java.language.Classifier
-import io.lionweb.lioncore.java.language.Concept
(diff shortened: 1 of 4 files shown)


lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -3,40 +3,20 @@ package com.strumenta.kolasu.lionweb
...
-import com.strumenta.kolasu.model.PossiblyNamed
-import com.strumenta.kolasu.model.ReferenceByName
-import com.strumenta.kolasu.model.allFeatures
-import com.strumenta.kolasu.model.asContainment
-import com.strumenta.kolasu.model.assignParents
-import com.strumenta.kolasu.model.isAttribute
-import com.strumenta.kolasu.model.isContainment
-import com.strumenta.kolasu.model.isReference
-import com.strumenta.kolasu.model.nodeOriginalProperties
+import com.strumenta.kolasu.model.*
```
ftomassetti, line 6: I thought ktlint would not like that. I configured my IDEA to avoid grouping imports in this way.
It is strange that the build is passing

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -183,16 +163,26 @@ class LionWebModelConverter(
...
-                                    lwNode.setReferenceValues(
-                                        StarLasuLWLanguage.ASTNodeOriginalNode,
-                                        listOf(
-                                            ReferenceValue(ProxyNode(targetID), null)
+                                val origin = kNode.origin
+                                if (origin is KNode) {
+                                    val targetID = myIDManager.nodeId(origin)
+                                    setOriginalNode(lwNode, targetID)
+                                } else if (origin is MissingASTTransformation) {
+                                    if (lwNode is DynamicNode) {
```
ftomassetti, line 171: Perhaps here we can check if it extends `AbstractClassifierInstance`, as it is a weaker request

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -307,6 +307,12 @@ class LionWebModelConverter(
                     require(originalNodeID != null)
                     referencesPostponer.registerPostponedOriginReference(kNode, originalNodeID)
                 }
+                val placeholderNodeAnnotation = lwNode.annotations.find {
```
ftomassetti, line 310: We should probably add a convenience method for this

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StarLasuLWLanguage.kt
```diff
@@ -21,6 +21,29 @@ object StarLasuLWLanguage : Language("com.strumenta.StarLasu") {
...
+            keyForContainedElement(PLACEHOLDER_NODE)
+        )
+        placeholderNodeAnnotation.annotates = LionCore.getConcept()
+        val reference =
+            Reference().apply {
+                this.name = "originalNode"
+                this.id = "${placeholderNodeAnnotation.id!!.removeSuffix("-id")}-$name-id"
+                this.key = "${placeholderNodeAnnotation.key!!.removeSuffix("-key")}-$name-key"
+                this.type = astNode
+                this.setOptional(true)
```
ftomassetti, line 42: I think it should not be optional because if the annotation is present, it should always refer to exactly one node that this is acting as a placeholder for

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StarLasuLWLanguage.kt
```diff
@@ -43,4 +66,10 @@ object StarLasuLWLanguage : Language("com.strumenta.StarLasu") {
 
     val char: PrimitiveType
         get() = StarLasuLWLanguage.getPrimitiveTypeByName("Char")!!
+
+    val PlaceholderNode: Annotation
+        get() = StarLasuLWLanguage.elements.filterIsInstance<Annotation>().find { it.name == PLACEHOLDER_NODE }!!
```
ftomassetti, line 71: We should probably add `Language.getAnnotationByName`

lionweb/src/test/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverterTest.kt
```diff
@@ -840,6 +827,44 @@ class LionWebModelConverterTest {
...
+        }
+        val mc = LionWebModelConverter()
+        mc.exportLanguageToLionWeb(kl)
+
+        val node1 = SimpleRoot(1, mutableListOf()).setSourceForTree(LionWebSource("MySource1"))
+        val node2 = SimpleRoot(2, mutableListOf()).setSourceForTree(LionWebSource("MySource2"))
+        node1.origin = MissingASTTransformation(node2)
+        val lwNode1 = mc.exportModelToLionWeb(node1)
+        // We verify the exported data is correct
+        val lwNode1Origins = lwNode1.getReferenceValueByName("originalNode")
```
ftomassetti, line 844: I was not expecting that: I thought the reference was on the annotation, not on the note itself

ftomassetti: OK, sorry, both the annotation and the node have two references called "originalNode"

ftomassetti:
It looks good to me, I have just minor comments

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
alessiostalla: I thought I had configured IDEA like that too... however I recently updated it and can't find the option anymore (now apparently Kotlin is built into the IDE and it's no longer a separate plugin)

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StarLasuLWLanguage.kt
alessiostalla: It's optional because:
1. the origin is optional in Kolasu as well
2. the origin could be something that cannot be represented in Lionweb, such as a parse tree node

lionweb/src/test/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverterTest.kt
alessiostalla: Yes, but the annotation's one is not used. I used a trick because in general, while annotations can have references in Lionweb, Kolasu's reference resolver only works on Nodes. So while in Kolasu we have `node --> missing_origin --> origin`, in Lionweb I represented this as `node with annotation --> origin`.

ftomassetti (approved):
Thank you @alessiostalla . For me it is good, please go ahead and merge

ftomassetti:
I have just released Kolasu 1.5.60 to include this PR and #356

PR #360, opened 2024-06-27
Closes #359 - Adds an ASTParserInstantiator interface relaxing the constraints imposed by KolasuParserInstantiator

### LINKED ISSUE #359: Relax constraints for Kolasu parser instantiators
When instantiating a parser, a corresponding instantiator is used at the moment. In order to implement a custom one, we can implement the `KolasuParserInstantiator` interface and bind it in our starlasu-conf.yaml file.

The current `KolasuParserInstantiator` can only be used with parsers extending `KolasuParser` and it would be nice to relax this constraint to `ASTParser`. For example, this would allow to use parser instantiators in the RPG parser project - where the main parser does not extend `KolasuParser`.

Using `ASTParser` instead of `KolasuParser` would still work as no methods or property specific of this last one is currently used. Indeed, we invoke the `parse` method on the instantiated parser and this method is inherited from `ASTParser`.

A possible solution to avoid breaking existing instantiators could consist in introducing an `ASTParserInstantiator` interface and make the `KolasuParserInstantiator` extend it.

Also see: https://github.com/Strumenta/starlasu-tools/issues/106

loradd:
Closes #359 - Adds an ASTParserInstantiator interface relaxing the constraints imposed by KolasuParserInstantiator

This PR introduces the following changes:
- Introduces an `ASTParserInstantiator` interface used to create `ASTParser` instances
- Modifies the existing `KolasuParserInstantiator` interface to extend the newly introduced one

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/parsing/ASTParserInstantiator.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/parsing/ASTParserInstantiator.kt
@@ -1,0 +1,13 @@
+package com.strumenta.kolasu.parsing
+
+import java.io.File
+
+/**
+ * This interface is used when need to configure parsers. For example, it can be used for testing, when we need
+ * to instantiate a parser with a particular configuration (e.g. specifying include lists of files).
+ * With respect to [KolasuParserInstantiator], this interface can be used to create more generic [ASTParser].instances.
+ **/
+interface ASTParserInstantiator {
+    fun instantiate(fileToParse: File): ASTParser<*>
+    fun instantiate(code: String): ASTParser<*>
+}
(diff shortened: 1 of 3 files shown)


loradd:
The build is currently failing because of some recent changes to the build.gradle.kts files related to ossh. I experienced this locally as well and temporarily fixed by replacing `property()` with `findProperty()` but I am not sure this is the final solution. 

Any additional information on this?

ftomassetti:
> The build is currently failing because of some recent changes to the build.gradle.kts files related to ossh. I experienced this locally as well and temporarily fixed by replacing `property()` with `findProperty()` but I am not sure this is the final solution.
> 
> Any additional information on this @ftomassetti ?

Yes, now for publishing we use the properties `ossrhTokenUsername` and `ossrhTokenPassword` instead of `ossrhUsername` and `ossrhTokenPassword`. However they should be used only for publishing, and if they are not present we should still be able to build. So perhaps your change to `findProperty` could make sense. Can you add it to this PR?

loradd:
Replaced `property` with `findProperty` in the semantics module. I am not sure why that is the only module causing problems at the moment.

Will merge as soon as accepted.

ftomassetti (approved):

PR #361, opened 2024-07-03
Define more node types as marker interfaces

alessiostalla:


ftomassetti (approved):
We discussed in a meeting. I would just consider revising the documentation of `TypeAnnotation`

PR #365, opened 2024-07-04
Integrate the KolasuClient in Kolasu

ftomassetti:
This component was previously part of a separate project. With this PR we move it into Kolasu. While doing so, we rename several packages.

Also, we created a base class to define Functional Tests using the LionWeb Repository


lionwebrepo-client/src/main/kotlin/com/strumenta/kolasu/lionwebclient/Utils.kt
```diff
@@ -0,0 +1,47 @@
...
+}
+
+fun LWNode.setParentID(parentID: String?) {
+    val parent =
+        if (parentID == null) {
+            null
+        } else {
+            ProxyNode(parentID)
+        }
+    (this as DynamicNode).parent = parent
```
alessiostalla, line 26: Shouldn't this method be defined on `DynamicNode` then?

alessiostalla (approved):
Looks good!

lionwebrepo-client/src/main/kotlin/com/strumenta/kolasu/lionwebclient/Utils.kt
ftomassetti: You are right. I will change it to work on `HasSettableParent` and not on `LWNode`

PR #370, opened 2024-07-10
Export StarLasu objects to Lionweb

alessiostalla:
This is a draft open for discussion.
The Kolasu-Lionweb integration is centered on Nodes. However Issues and the other non-Node classes are obviously not nodes. In the CIS they're made into nodes but we can't use the same method here (unless we port `BaseNode` and related stuff into Kolasu).
So this PR is a possible approach. Open issues include:
- the ID to assign to Issues and other exported objects. Currently I bailed out and made it into a parameter for the client to provide, but this only moves the problem elsewhere.
- the creation of specific methods one has to use when exporting said objects. E.g. `exportIssueToLionweb` in addition to `exportModelToLionWeb`


ftomassetti:
A review is requested, but the PR is in draft and we discussed this the other day in a meeting. So far everything seems to be going in the right direction, and I assume that a more detailed review is not needed at this time. I am writing this just to be sure I am not blocking or slowing you down

alessiostalla:
This is now in a good enough shape that it can be reviewed.
The approach has been switched to using lionweb-kotlin to represent Issue and ParsingResult so that it's no longer necessary to provide an ID immediately when exporting such objects to Lionweb. Additionally, ParsingResult LW nodes know how to compute their own ID from their Source, if it's set.

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebLanguageConverter.kt
```diff
@@ -60,6 +64,17 @@ class LionWebLanguageConverter {
         registerMapping(PlaceholderElement::class, StarLasuLWLanguage.PlaceholderElement)
         registerMapping(Statement::class, StarLasuLWLanguage.Statement)
         registerMapping(TypeAnnotation::class, StarLasuLWLanguage.TypeAnnotation)
+
+        registerMapping(Issue::class, StarLasuLWLanguage.Issue)
+        classesAndEnumerations.associate(
+            IssueSeverity::class,
+            (StarLasuLWLanguage.Issue.getFeatureByName("severity") as Property).type as Enumeration
```
ftomassetti, line 71: I would consider either creating constants for these features, or alternatively to use Issue::severity.name and Issue::type.name, as a way to ensure that if we change the name, this value is updated

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -702,11 +746,81 @@ class LionWebModelConverter(
...
+        return StarLasuLWLanguage.Issue
+    }
+}
+
+class ParsingResultNode(val source: Source?) : BaseNode() {
+    override fun calculateID(): String? {
+        try {
+            return SimpleSourceIdProvider().sourceId(source) + "_ParsingResult"
+        } catch (_: SourceShouldBeSetException) {
+            // return null
```
ftomassetti, line 818: Should this be uncommented?

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -702,11 +746,81 @@ class LionWebModelConverter(
         return kNode
     }
 
+    private fun <T : Any> maybeInstantiateSpecialObject(kClass: KClass<T>, data: Node): T? {
```
ftomassetti, line 749: I would consider adding a command for this

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -702,11 +746,81 @@ class LionWebModelConverter(
...
+        return when (kClass) {
+            Issue::class -> {
+                Issue(
+                    attributeValue(data, data.classifier.getPropertyByName("type")!!) as IssueType,
```
ftomassetti, line 753: I would cosider using Issue::type.name and similar values, to protect against typos

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
ftomassetti, line 818: Is this intended?

ftomassetti (approved):
Looks very good to me. I just added minor comments.

Please let me know if you need me to do a new release

alessiostalla:
I don't _need_ a release right now, but if you release it, I'll take a shot at making use of this in the CIS.

ftomassetti:
I tried making a release but Sonatype seems to keep going in timeout today

PR #382, opened 2024-09-03
Starlasu classes in Lionweb

alessiostalla:
Include tokens list in _ParsingResult_ instances serialized to Lionweb


core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
```diff
@@ -82,7 +82,7 @@ data class TokenCategory(val type: String) {
 open class KolasuToken(
     open val category: TokenCategory,
     open val position: Position,
-    open val text: String
+    open val text: String? = null
```
ftomassetti, line 85: Do we want to allow this? Why should a token have no text?

If we keep it this way, I would consider adding a comment to explain under which circumstances `text` can be null

core/src/test/kotlin/com/strumenta/kolasu/parsing/KolasuParserTest.kt
```diff
@@ -51,7 +51,7 @@ class KolasuParserTest {
             """.trimMargin()
         )
         assertNotNull(result)
-        val lexingResult = parser.tokenFactory.extractTokens(result)
+        val lexingResult = (parser.tokenFactory as ANTLRTokenFactory).extractTokens(result)
```
ftomassetti, line 54: Should the `extractTokens` method be exposes by TokenFactory to avoid this cast?

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -82,6 +84,22 @@ class LionWebModelConverter(
...
+        private val kFeaturesCache = mutableMapOf<Class<*>, Map<String, Feature>>()
+        private val lwFeaturesCache = mutableMapOf<Classifier<*>, Map<String, LWFeature<*>>>()
+
+        fun lwFeatureByName(classifier: Classifier<*>, featureName: String): LWFeature<*>? {
+            return lwFeaturesCache.getOrPut(classifier) {
+                classifier.allFeatures().associateBy { it.name!! }
+            }[featureName]
+        }
+
+        init {
```
ftomassetti, line 97: Nice!

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -740,11 +747,13 @@ class LionWebModelConverter(
             }
             ParsingResult::class -> {
                 val root = data.getOnlyChildByContainmentName(ParsingResult<*>::root.name)
-                ParsingResult(
+                val tokens = data.getPropertyValue(data.classifier.getPropertyByName("tokens")!!) as TokensList?
```
ftomassetti, line 750: Instead of a string, when possible I use a property reference. Like `ParsingResult::tokens.name`. It helps to find usages and protect against typos

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
```diff
@@ -763,26 +772,41 @@ class LionWebModelConverter(
 
     fun exportIssueToLionweb(issue: Issue): IssueNode {
         val issueNode = IssueNode()
-        issueNode.setPropertyValue(StarLasuLWLanguage.Issue.getPropertyByName("message")!!, issue.message)
-        issueNode.setPropertyValue(StarLasuLWLanguage.Issue.getPropertyByName("position")!!, issue.position)
-        setEnumProperty(issueNode, StarLasuLWLanguage.Issue.getPropertyByName("severity")!!, issue.severity)
-        setEnumProperty(issueNode, StarLasuLWLanguage.Issue.getPropertyByName("type")!!, issue.type)
+        issueNode.setPropertyValue(StarLasuLWLanguage.Issue.getPropertyByName(Issue::message.name)!!, issue.message)
```
ftomassetti, line 775: Good!

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StarLasuLWLanguage.kt
```diff
@@ -63,10 +66,12 @@ object StarLasuLWLanguage : Language("com.strumenta.StarLasu") {
             addProperty("position", position, Multiplicity.OPTIONAL)
         }
 
+        addPrimitiveType(TokensList::class)
         ParsingResult = addConcept("ParsingResult").apply {
             addContainment("issues", Issue, Multiplicity.MANY)
             addContainment("root", ASTNode, Multiplicity.OPTIONAL)
```
ftomassetti, line 72: Also here, if we could use property references for names it could be slightly more robust

ftomassetti (approved):
It looks good to me. Thank you @alessiostalla

core/src/main/kotlin/com/strumenta/kolasu/parsing/Parsing.kt
alessiostalla: When storing them in the LW repo, we don't include the text

core/src/test/kotlin/com/strumenta/kolasu/parsing/KolasuParserTest.kt
alessiostalla: No because it used to be exposed by `TokenFactory` but internally it assumed ANTLR tokens which is incorrect. That's just a test, user code is not supposed to do that.

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/StarLasuLWLanguage.kt
```diff
@@ -184,4 +189,33 @@ private fun registerSerializersAndDeserializersInMetamodelRegistry() {
         positionSerializer,
         positionDeserializer
     )
+
+    val tokensListPrimitiveSerializer = PrimitiveSerializer<TokensList?> { value: TokensList? ->
+        value?.tokens?.joinToString(";") { kt ->
```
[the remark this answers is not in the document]
alessiostalla: I would prefer to release this as soon as possible, so we can test the integration in the CIS, and only then optimize it.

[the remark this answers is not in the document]
ftomassetti: Makes sense

lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/LionWebModelConverter.kt
alessiostalla: I've updated the codebase to use property references here and in other places

ftomassetti:
@alessiostalla let me know if you need me to do a release right away or if you prefer to do so yourself

PR #383, opened 2024-09-17
Fault tolerant transformers

ftomassetti:
When we implement transformers we may have the case in which the transformation is missing and the case in which it is failing. It may be failing because we cover a certain subset of cases. For example, when transforming from RPG to Java, we may cover a field declaration if the field is a string or a number, but not if it is of type pointer. In these cases we may want to still produce a partial transformation, treating the failure similarly to what we do with missing transformations. So expand the strategy for missing transformation to cover also failing transformations.

In addition to this:
* We ensure that generated IDs are compatible with LionWeb, removing illegal characters
* We expand the logic to instantiate a dummy node, with the sole purpose of attaching an origin to it. Previously we were needing a default constructor. Now, if that is not present, we attempt to use other constructors, generating dummy values as needed

### DIFF
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
@@ -12,6 +12,12 @@ import kotlin.reflect.full.isSubclassOf
 import kotlin.reflect.full.primaryConstructor
 import kotlin.reflect.jvm.javaType
 
+/**
+ * This logic instantiate a node of the given class with dummy values.
+ * This is useful because it permits to add an element that "fit" and make the typesystem happy.
+ * Typically, the only goal of the element would be to hold some annotation that indicates that the element
+ * is representing an error or a missing transformation or something of that sort.
+ */
 fun <T : Node> KClass<T>.dummyInstance(): T {
     val kClassToInstantiate = this.toInstantiableType()
     val emptyConstructor = kClassToInstantiate.constructors.find { it.parameters.isEmpty() }
@@ -77,6 +83,10 @@ private fun <T : Any> KClass<T>.toInstantiableType(): KClass<out T> {
     }
 }
 
+/**
+ * We can only instantiate concrete classes or sealed classes, if one its subclasses is directly or
+ * indirectly instantiable. For interfaces this return false.
+ */
 fun <T : Any> KClass<T>.isDirectlyOrIndirectlyInstantiable(): Boolean {
     return if (this.isSealed) {
         this.sealedSubclasses.any { it.isDirectlyOrIndirectlyInstantiable() }
--- a/core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
+++ b/core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
@@ -3,8 +3,9 @@ package com.strumenta.kolasu.transformation
 import com.strumenta.kolasu.model.Node
 import com.strumenta.kolasu.model.PossiblyNamed
 import com.strumenta.kolasu.model.ReferenceByName
+import com.strumenta.kolasu.model.isContainment
+import com.strumenta.kolasu.model.nodeProperties
 import java.lang.reflect.ParameterizedType
-import kotlin.random.Random
 import kotlin.reflect.KClass
 import kotlin.reflect.KParameter
 import kotlin.reflect.full.createType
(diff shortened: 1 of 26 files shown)


core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
```diff
@@ -0,0 +1,96 @@
...
+    val params = mutableMapOf<KParameter, Any?>()
+    constructor.parameters.forEach { param ->
+        val mt = param.type.javaType
+        val value = when {
+            param.type.isMarkedNullable -> null
+            mt is ParameterizedType && mt.rawType == List::class.java -> mutableListOf<Any>()
+            (param.type.classifier as KClass<*>).isSubclassOf(Node::class) ->
+                (param.type.classifier as KClass<out Node>).dummyInstance()
+            param.type == String::class.createType() -> "DUMMY"
+            param.type.classifier == ReferenceByName::class -> ReferenceByName<PossiblyNamed>("UNKNOWN")
```
alessiostalla, line 37: This is unlikely, but the reference could be "resolved" to a real element called "UNKNOWN". Is it possible to have a _ReferenceByName_ subclass that can't possibly be resolved?

core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
```diff
@@ -0,0 +1,96 @@
...
+            val subClassWithEmptyParam = subclasses.find { it.constructors.any { it.parameters.isEmpty() } }
+            if (subClassWithEmptyParam == null) {
+                if (subclasses.size == 1) {
+                    subclasses.first().toInstantiableType()
+                } else {
+                    // Some constructs are recursive (think of the ArrayType)
+                    // We either find complex logic to find the ones that aren't or just pick one randomly.
+                    // Eventually we will build a tree
+                    val r = Random.Default
+                    subclasses[r.nextInt(subclasses.size)].toInstantiableType()
```
alessiostalla, line 68: I'm not a fan of random here because parsing the same file twice will likely give different results, which I think is undesirable (even if we can say that the behavior is unspecified for code that we don't yet support).
I would also consider a mechanism (marker interface or annotation) with which we can designate a specific subclass as the default/fallback node, because in my experience it's not uncommon to have classes such as _UnsupportedStatement_ or _ExpressionError_ already in use.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
```diff
@@ -578,7 +607,7 @@ open class ASTTransformer(
         return nodeFactory
     }
 
-    fun <T : Node> registerIdentityTransformation(nodeClass: KClass<T>) =
+    inline fun <reified T : Node> registerIdentityTransformation(nodeClass: KClass<T>) =
```
alessiostalla, line 610: What's the reason for inline+reified here?

core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
ftomassetti: That is a good point. My goal is to avoid that we pick always the wrong one as `ArrayType(ArrayType(ArrayType(...)))`, but perhaps what we can do is to pick a pseudo-random number, that actually depends on the depth of the "DummyNode". I will work on that. Adding an annotation could make our life easier here, but it would require extra work when building the language, so I would lean toward doing more work on this code to have to do less work on all the languages

core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
ftomassetti: I see your point, but I would aim to solve it by never trying to resolve references belonging to "dummy nodes" or "dummy trees". So I would consider:
1. Adding a flag `val Node.isDummy` that looks at the origin parameter and possibly also to the parent
2. Adding a flag in `walk` and similar methods, to give the option to skip dummy nodes
3. When resolving references use the `walk` method using the flag that skip dummy nodes
In this way we never attempt to solve references. Alternatively we can also create subclass of ReferenceByName that explode if you attempt to mark it as solved. 
What do you think?

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
ftomassetti: That is a good question. I think I needed to access the node type in some method invoked from here and the IDE suggested this refactoring.

In particular I think it was due to this other change:
```
inline fun <S : Any, reified T : Node> registerNodeFactory(
```
this became inlines and T parameter became reified, so that there we could write: `val node = T::class.dummyInstance()`. But given `registerIdentityTransformation` call `registerNodeFactory`, that had to be changed too. Perhaps I can add a comment about that.

core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
alessiostalla: I'm not a fan of modifying _walk_, but the symbol resolver could apply a filter to the nodes it walks to exclude dummy ones, it makes sense

core/src/main/kotlin/com/strumenta/kolasu/transformation/DummyNodes.kt
alessiostalla: I see. Anyway I guess this only applies to 1.5 because in 1.6 we'll have machine-generated error nodes to use.

core/src/main/kotlin/com/strumenta/kolasu/transformation/Transformation.kt
alessiostalla: Ok, thanks.

ftomassetti:
Thank you @alessiostalla for the comments, I think the PR is now improved

PR #388, opened 2025-01-31
Propagate expected type

ftomassetti:
There are circumstances in which we use the expected type to instantiate dummy nodes. For this to work we need to propagate the expected type in different circumstances.


ftomassetti:
hi @alessiostalla , would you have time to review this?

alessiostalla (approved):

ftomassetti:
Thank you @alessiostalla !

PR #390, opened 2025-02-05
Revert to use LW 2023.1 format

ftomassetti:
Depends on:
https://github.com/LionWeb-io/lionweb-java/pull/179
https://github.com/LionWeb-io/lionweb-kotlin/pull/8

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -4,8 +4,8 @@ superPublish = { id = "com.vanniktech.maven.publish", version = "0.22.0" }
 ktlint = { id = "org.jlleitschuh.gradle.ktlint", version = "11.5.1" }
 
 [versions]
-lwjava = "0.3.0"
-lwkotlin = "0.3.0"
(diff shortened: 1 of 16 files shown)


ftomassetti:
The build is currently failing because LW Kotlin is currently propagating to Maven Central. Restarting the build in a few minutes should make it green.

core/src/main/kotlin/com/strumenta/kolasu/mapping/Support.kt
gabriele-tomassetti, line 1:
```
return original?.let { transform(it, expectedType = T::class) as? T }
```
If am not mistaken you could also use `as?` to handle the case when the cast fails.

ftomassetti: Yes, but I want to throw an exception if the type is not the one expected. While null is fine instead

gabriele-tomassetti (approved):

PR #391, opened 2025-02-06
Bugfix for identity transformation

ftomassetti:
So far an identity transformation would keep the original node. The problem is that in this way we do not apply transformations to its descendants.

With this PR we change approach: the identity transformation makes a copy of the node, applying transformations to che children. In this way we can define a few transformations and transform only certain nodes, while keeping other nodes as in the original AST (besides their children being transformed).


tiagobstr (approved):

tiagobstr:
Looks good @ftomassetti

ftomassetti:
Thank you Tiago!

PR #393, opened 2025-02-11
fixing identity origin + adding test to check it

tiagobstr:
Adding origin to the identity mapping and a test to confirm it

### DIFF
--- a/core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
+++ b/core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
@@ -503,8 +503,30 @@ class ASTTransformerTest {
             transformer1.transform(original) as AA
         )
     }
-}
 
+    @Test
+    fun testIdentityTransformationOfIntermediateNodesWithOrigin() {
+        val transformer1 = ASTTransformer(defaultTransformation = IDENTTITY_TRANSFORMATION)
+        transformer1.registerNodeFactory(AA::class) { original, t, _ ->
+            BA("your_" + original.a.removePrefix("my_"), t.translateCasted(original.child))
+        }
+        val original = AA(
+            a = "my_a",
+            child = AB(
+                b = "my_b",
+                child = AC(
+                    c = "my_c",
+                    children = mutableListOf(
+                        AD("my_d1")
+                    )
+                )
+            )
+        )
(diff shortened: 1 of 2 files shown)


core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
```diff
@@ -503,8 +503,30 @@ class ASTTransformerTest {
...
+                b = "my_b",
+                child = AC(
+                    c = "my_c",
+                    children = mutableListOf(
+                        AD("my_d1")
+                    )
+                )
+            )
+        )
+        val transformedAST = transformer1.transform(original) as AA
```
ftomassetti, line 525: I think it would be more obvious what is going on if we casted to BA. This works as BA extends AA but it took me a while to understand it

core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
```diff
@@ -503,8 +503,30 @@ class ASTTransformerTest {
...
+        // verify that the origin is set correctly
+        assert(transformedAST.origin == original)
```
ftomassetti, line 527: I would use assertEquals and not assert as assertEquals produce a better error message and assert may be skipped depending on the compilation flags, I think

I would consider also checking that all the descandants nodes have the corresponding origin.

I think we can do something llke:
assertEquals(transformedAST.walkDescendants(AB::class).first().origin, original.walkDescendants(AB::class).first())

And similary for AC and AD

ftomassetti (approved):
I just added a couple of suggestions, but the code seems correct to me

core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
tiagobstr: Makes sense, thanks!

core/src/test/kotlin/com/strumenta/kolasu/transformation/ASTTransformerTest.kt
tiagobstr: Makes sense, thanks!

ftomassetti (approved):

PR #394, opened 2025-03-03
Consider only original properties in Node.toString

ftomassetti:
We should not try to print also derived properties as they could lead to redundant information being printed, and possible circular structures (leading to stack overflows).


alessiostalla (approved):

ftomassetti:
Thank you @alessiostalla !

PR #398, opened 2025-04-09
Adopting Starlasu Specs

ftomassetti:
We move the StarlasuLWLanguage to Starlasu Specs, so that it can be easily shared across platforms and it can be kept stable, changing less frequently than Kolasu does.

Todo:

- [x] Migrate remaining references to StarLasuLWLanguage, outside of StarLasuLWLanguage itself, to ASTLanguage
- [x] Do not touch StarLasuLWLanguage itself
- [x] Move primitive serializers and deserializers to a separate class

**Note**: some tests may be failing because of dependencies still propagating to Maven Central. Eventually those tests should pass


ftomassetti:
This change is a pre-requisite for a PR on the CodeInsightStudio. It shows the direction of moving the "standard" LionWeb languages at the basis of the Starlasu libraries to [Starlasu-Specs](https://github.com/strumenta/starlasu-specs). Starlasu-Specs would define the "contract" that the different Starlasu components should respect.

tiagobstr:
After looking into it and trying it, looks good to me.

tiagobstr (approved):

ftomassetti:
Thank you @tiagobstr !

PR #402, opened 2025-04-18
Update LionWeb dependencies

ftomassetti:
As part of this PR we do minor maintenance operations:
- We update the version of gson used and adopt the version catalog for managing such version
- We update the versions of LW Java, LW Kotlin,  StarLasu Specs
- We add the parameter jsonSerialization to convertCodebase
- We expose the serializers and deserializers for the StarLasuLWLanguage (a.k.a. ASTLanguage), so that they can be used individually from users of Kolasu
- We update TodoFunctionalTest to create a separate repository for each test. This is necessary because now the base class AbstractRepoClientFunctionalTest does not recreate the whole DB for each test
- In KolasuClient we do not keep a separate JsonSerialization, but we use the one of the wrapped LionWebClient

Depends on https://github.com/LionWeb-io/lionweb-kotlin/pull/9

### DIFF
--- a/codebase/build.gradle.kts
+++ b/codebase/build.gradle.kts
@@ -10,7 +10,6 @@ plugins {
 
 val jvmVersion = project.property("jvm_version") as String
 val kotlinVersion = project.property("kotlin_version") as String
-val gsonVersion = project.property("gson_version") as String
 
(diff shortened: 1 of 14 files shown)


tiagobstr (approved):
Looks good to me .

ftomassetti:
Thank you @tiagobstr

PR #403, opened 2025-05-05
Fix serialization

ftomassetti:
Fixing minor issues emerged while serializing to LionWeb.

* We update the Starlasu Specs to 0.2.1
* We ensure that when calculating the original properties we do not first calculate all properties (including derived properties) to later filter them out. This is important as calculation of derived properties could fail while we are building the node
* We correct support for serializing and deserializing null values for primitive types

### DIFF
--- a/javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
+++ b/javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
@@ -4,6 +4,7 @@ import com.strumenta.kolasu.model.*;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
 
@@ -16,17 +17,16 @@ public class CompilationUnit extends Node {
             return bs;
         }
 
-        @NotNull
         @Override
-        @Internal
-        public List<PropertyDescription> getProperties() {
+        @NotNull
+        public List<PropertyDescription> getOriginalProperties() {
             return Arrays.asList(new PropertyDescription("bs", true, Multiplicity.MANY, getBs(), PropertyType.CONTAINMENT, false));
         }
 
         @Override
         @NotNull
-        public List<PropertyDescription> getOriginalProperties() {
-            return getProperties();
+        public List<PropertyDescription> getDerivedProperties() {
+            return Collections.emptyList();
         }
     }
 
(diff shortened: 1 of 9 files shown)


lionweb/src/main/kotlin/com/strumenta/kolasu/lionweb/PrimitiveSerialization.kt
```diff
@@ -63,19 +67,21 @@ val pointDeserializer: PrimitiveDeserializer<Point> =
...
-        val parts = serialized.split("-")
-        require(parts.size == 2) {
-            "Position has an expected format: $serialized"
-        }
-        Position(pointDeserializer.deserialize(parts[0]), pointDeserializer.deserialize(parts[1]))
+        return@PrimitiveDeserializer null
+    }
+    val parts = serialized.split("-")
+    require(parts.size == 2) {
+        "Position has an expected format: $serialized"
```
alessiostalla, line 82: Maybe you meant "unexpected"?

javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
```diff
@@ -22,6 +22,12 @@ public List<B> getBs() {
         public List<PropertyDescription> getProperties() {
             return Arrays.asList(new PropertyDescription("bs", true, Multiplicity.MANY, getBs(), PropertyType.CONTAINMENT, false));
         }
+
+        @Override
+        @NotNull
+        public List<PropertyDescription> getOriginalProperties() {
```
alessiostalla, line 28: Is it necessary to always override this method in Java? Or only when overriding _getProperties_?

alessiostalla (approved):

javalib/src/test/java/com/strumenta/kolasu/javalib/CompilationUnit.java
ftomassetti: It is currently always required, let me change that

ftomassetti:
Functional tests currently fail because of an issue with a docker image of the LionWeb Repository. This is due to the fact that in that build we build the lionweb repository and its build started to fail as some dependency was not specifically connected to a version, so when running now it gets the latest version and it fails to build. We will report that, but in the meantime let's release Kolasu anyway.

ftomassetti:
The issue has been reported as https://github.com/LionWeb-io/lionweb-repository/issues/116

PR #404, opened 2025-07-16
Update to LW Java 1.0.1

ftomassetti:


ftomassetti:
> Everything looks in order!

Thank you! I have also updated to a slightly newer versions of LW Java and LW Kotlin, with performance optimizations

PR #406, opened 2025-07-23
Update deps

ftomassetti:
Most changes are due to Ktlint reformatting the files. I am sorry for the noise and I am not sure why ktlint acted up: I had to remove the old configuration as I was getting an error but I did not updated it (however I removed the configuration forcing an old version of the ktlint engine to be used)


ftomassetti:
I will not require a review as this PR is too noisy

